PR #2, opened 2021-10-14
APS-372 create initial type-of-party page

pk2453:


packages/forms-web-app/src/views/register/type-of-party.njk
```diff
@@ -0,0 +1,72 @@
+{% extends "layouts/main.njk" %}
```
elvin-ali, line 1: @pk2453 @AnthonyJWalsh Is it acceptable for the page **header** to _look_ different from design  ? Shall we raise additional ticket ( as this is something generic for the whole app as opposed to being specific to this ticket ) ? 
VS

lvnhmd (approved):

PR #7, opened 2021-10-19
APS-404 add guidence pages

pk2453:


tom-f-difrent (approved):
Would have liked a bit more time to discuss, but we are pusshed for time. But all in order, you stept up to the plate. Obviously some tech debt here but for another time.

PR #22, opened 2021-11-01
Add trialists list

lvnhmd:


AnthonyJWalsh (changes requested):
LGTM! This is exactly what I was thinking :)

We will however need to add the following line to the applications-service-api Dockerfile

`ADD data data
`
And ensure that we add a TRIALIST_DATA_PATH config setting to the dev environment (which I will do now)

AnthonyJWalsh (approved):

PR #38, opened 2021-11-17
acceptance tests for 454 and 426 tickets

harish-kancharla:


pk2453 (commented):
Not sure if you should be using hardcoded project name and case reference as they are dynamic and could change. Your approach should be from project search-page and select the desired link from there to go overview page and then to document library page.

pk2453 (approved):

PR #83, opened 2022-01-05
Implement APS-574 and APS-662

lvnhmd:


pk2453 (commented):
I think now we have 2 separate implementation for same thing. (display closure date in string format) here I have already implemented this https://github.com/Planning-Inspectorate/applications-service/blob/main/packages/forms-web-app/src/controllers/overview.js#L22 so may be putting in some utility file and reuse?

lvnhmd:
> I think now we have 2 separate implementation for same thing. (display closure date in string format) here I have already implemented this https://github.com/Planning-Inspectorate/applications-service/blob/main/packages/forms-web-app/src/controllers/overview.js#L22 so may be putting in some utility file and reuse?

pk2453 (approved):

PR #86, opened 2022-01-06
Fix logic issue around Save and Exit for org and agent

lvnhmd:


pk2453 (changes requested):
I this logic need to be checked and modified for other 2 journeys i.e. self and organisation

lvnhmd:
> I this logic need to be checked and modified for other 2 journeys i.e. self and organisation

This PR modifies the logic for org (packages/forms-web-app/src/controllers/register/organisation/comments.js) and agent (packages/forms-web-app/src/controllers/register/behalf/comments.js)

The logic for self is already in place ( https://github.com/Planning-Inspectorate/applications-service/blob/f1c656c7c82a829f5e60fe0858de0816d4b0d8e4/packages/forms-web-app/src/controllers/register/myself/comments.js#L26 )

pk2453 (approved):
If logic for other journey is in place then good to go

PR #98, opened 2022-01-17
APS-702-709-712-acceptance-tests

harish-kancharla:


e2e-tests/cypress/integration/are-you-sure-you-want-to-remove-this-comment.feature
```diff
@@ -1,4 +1,4 @@
-@testSuite
+@ignore
```
pk2453, line 1: Is this still meant to be @ignore ?

e2e-tests/cypress/integration/do-you-want-to-add-another-comment.feature
```diff
@@ -1,4 +1,4 @@
-@testSuite
+@ignore
```
pk2453, line 1: same here

harish-kancharla: since this page doesn't exist anymore I have just ignored this feature file for the time being instead of deleting them because we never know it might come back 😄

PR #101, opened 2022-01-20
Add page capturing organisation name

lvnhmd:


pk2453:
URL may change as part of other ticket
Title needs to be updated as per ticket
No validation on this page, is this not required?
Dont work for organisation link takes to wrong page

pk2453:
URL may change as part of other ticket
Title needs to be updated as per ticket
No validation on this page, is this not required?
Dont work for organisation link takes to wrong page

lvnhmd:
URL may change as part of other ticket 
_Yes, I made a conscious decision to leave the url be changed as part of a common url consolidation  ticket_
Title needs to be updated as per ticket 
_Done_
No validation on this page, is this not required? 
_If we have validation enabled, we will not be able to skip via link **I don't work for an organisation**_
Dont work for organisation link takes to wrong page
_Fixed_

pk2453:
**If we have validation enabled, we will not be able to skip via link I don't work for an organisation**
I think validation will only come in picture if you click the button. But if you click the link validation should't come in picture.

lvnhmd:
> **If we have validation enabled, we will not be able to skip via link I don't work for an organisation** I think validation will only come in picture if you click the button. But if you click the link validation should't come in picture.

_Correct, thank you, just added validation back_

packages/forms-web-app/src/validators/register/behalf/organisation-name.js
```diff
@@ -0,0 +1,9 @@
+const { body } = require('express-validator');
+
+const rules = () => {
+  return [body('organisation-name').notEmpty().withMessage('Enter your organisation name')];
```
pk2453, line 4: Sorry, another change. Error message should be **Enter the name of the organisation you work for**

lvnhmd: Changed.

PR #102, opened 2022-01-21
Feature/aps 723

pk2453:


lvnhmd:
Register to have your say (landing page) page title not changed

pk2453:
> Register to have your say (landing page) page title not changed
This is fixed.

PR #106, opened 2022-01-31
ASB-9 After decision page changes

pk2453:


packages/forms-web-app/src/views/interested-party-guide/get-involved-preliminary-meetings.njk
```diff
@@ -5,13 +5,13 @@
 {% endblock %}
```
lvnhmd, line 5: It is the same with the font here as well. If you compare [https://applications-prototype.herokuapp.com/mvp/interested-party-guide/what-you-can-do-after-decision](https://applications-prototype.herokuapp.com/mvp/interested-party-guide/what-you-can-do-after-decision) and [http://localhost:9004/interested-party-guide/what-happens-after-decision](http://localhost:9004/interested-party-guide/what-happens-after-decision). I inspected the pages in the browser, but could not see difference. It is probably the same for the pages I worked on, but I notice it only now. Do we ignore this as display issue ?

lvnhmd (changes requested):

packages/forms-web-app/src/views/interested-party-guide/get-involved-preliminary-meetings.njk
lvnhmd: Would you please paste confirmation link ?

pk2453: Which confirmation link? https://design-system.service.gov.uk/styles/typography/ here is font style.

lvnhmd (approved):

PR #132, opened 2022-03-01
ASB-155: Made route to registration start parameterised with project case ref

cpcundill:


packages/forms-web-app/__tests__/unit/controllers/register/start.test.js
```diff
@@ -1,44 +1,103 @@
...
+            DateOfRelevantRepresentationClose: '2024-09-01',
+          },
+        })
+      );
+      const req = {
+        ...mockReq(),
+        params: {
+          case_ref: 'ABC123',
+        },
         session: {
```
pk2453, line 58: I think we don't need session any more and can be removed from here

cpcundill: Hey @pk2453 - in this scenario I kept the session to prove that any existing session value will not be used but rather the project details will be loaded based on the incoming case_ref in the URL path.  The scenario title is:

should load project data and return register start view even when existing project in session

pk2453 (approved):

PR #135, opened 2022-03-01
Add release pipelines for new environments

andylangridge:
Have added a couple of pipelines (CI and CD) which deploy to the new environments. Had to duplicate the existing Dockerfile, which I've moved temporarily into the pipelines folder.

Also added a pipeline for the common package. This is just to get things working for now. I think in future it would be good to remove this dependency in the Dockerfile, but for now I've replicated things kind of as they are now.

### DIFF
--- a/pipelines/common/common-ci.yml
+++ b/pipelines/common/common-ci.yml
@@ -1,0 +1,45 @@
+trigger:
+  branches:
+    include:
+      - main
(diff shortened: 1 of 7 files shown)


cpcundill (approved):
Happy for you to merge based on your explanation of the changes.  Thanks 👍

PR #136, opened 2022-03-02
Add env var PRIVATE_BETA_V1_ROUTES_ONLY

lvnhmd:


packages/forms-web-app/src/config.js
```diff
@@ -42,5 +42,6 @@ module.exports = {
   },
   featureFlag: {
     googleTagManager: process.env.FEATURE_FLAG_GOOGLE_TAG_MANAGER === 'true',
+    usePrivateBetaV1RoutesOnly: process.env.PRIVATE_BETA_V1_ROUTES_ONLY === 'true',
```
cpcundill, line 45: Default should be `false` since we'll only set it to be true for production

cpcundill (changes requested):

packages/forms-web-app/src/config.js
lvnhmd: The default is false, [ ](https://github.com/Planning-Inspectorate/applications-service/blob/a56860804c1e287632f29eae52696947dbdc140e/docker-compose.yml#L56)

PR #140, opened 2022-03-03
ASB-14: add pagination feature

pk2453:


packages/forms-web-app/src/routes/examination/about-the-application.js
```diff
@@ -3,8 +3,8 @@ const router = express.Router();
 
 const aboutTheApplicationController = require('../../controllers/examination/about-the-application');
 
-router.get('/:page', aboutTheApplicationController.getAboutTheApplication);
-router.post('/search/:page', aboutTheApplicationController.postSearchDocument);
+router.get('/:case_ref/:page', aboutTheApplicationController.getAboutTheApplication);
+router.post('/search/:case_ref/:page', aboutTheApplicationController.postSearchDocument);
```
cpcundill, line 7: Don't you think it would be better to have the `:case_ref` before the `/search` portion of the URL?  That would be consistent with how we link to the examination route in the first place, e.g. `/examination/EN010118`.  Thus, the search route would follow more logically as a sub-resource: `/examination/EN010118/about-the-application/search` or maybe better still: `/examination/EN010118/search`

cpcundill (changes requested):
Looking good generally  - just one comment so far.

packages/forms-web-app/src/routes/examination/about-the-application.js
pk2453: Its updated to **/examination/EN010118/about-the-application/search**

packages/forms-web-app/src/views/examination/about-the-application.njk
```diff
@@ -26,18 +26,22 @@
...
-                    <span id="results-number">{{ pageData.totalItems }}</span> results
+                    No documents found. Please try with different search text.
                   {% endif %}
                 </strong>
               </p>
             </div>
 
             <button id="toggleFilters" class="govuk-button govuk-button">Show search and filters</button>
-            <form action="/document-library/search/{{ caseRef }}/1" method="POST" novalidate>
+            <form action="/examination/about-the-application/{{ caseRef }}/search/1" method="POST" novalidate>
```
cpcundill, line 44: Should this link be hard-coded?  Looks like it's now out of step with the route change?

packages/forms-web-app/src/views/layouts/vertical-tabs.njk
```diff
@@ -16,7 +16,7 @@
     </li>
     <li class="side-menu govuk-body-s {% if 'about-the-application' in current %} active {% endif %}">
       {% if not 'about-the-application' in current %}
-        <a href="/examination/about-the-application/1">Project application documents</a>
+        <a href="/examination/about-the-application/{{ caseRef }}/1">Project application documents</a>
```
cpcundill, line 19: Again href looks inconsistent with route change.

cpcundill (changes requested):

packages/forms-web-app/src/routes/examination/about-the-application.js
```diff
@@ -3,8 +3,8 @@ const router = express.Router();
 
 const aboutTheApplicationController = require('../../controllers/examination/about-the-application');
 
-router.get('/:page', aboutTheApplicationController.getAboutTheApplication);
-router.post('/search/:page', aboutTheApplicationController.postSearchDocument);
+router.get('/:case_ref/:page', aboutTheApplicationController.getAboutTheApplication);
+router.post('/:case_ref/search/:page', aboutTheApplicationController.postSearchDocument);
```
cpcundill, line 7: @pk2453 Surely these routes need to be registered within `packages/forms-web-app/src/routes/examination/index.js` so that they don't contain `about-the-application` ?

pk2453: Well, its just one of way to write for easy navigation otherwise you can have all your routes in single file

packages/forms-web-app/src/views/examination/about-the-application.njk
pk2453: all routes/links updated to have /examination/{{case_ref}}

packages/forms-web-app/src/views/layouts/vertical-tabs.njk
pk2453: all routes/links updated to have /examination/{{case_ref}}

cpcundill (changes requested):
The point I was making about the routes is that we have an existing route `/examination/:case_ref`. We should scope the `about-the-application` sub-resource under that.  So we might as well mount `aboutTheApplicationRouter`  at `/examination/:case_ref/about-the-application`

cpcundill (approved):
Cheers for all your hard work on this 👍

PR #145, opened 2022-03-07
Do not delete session

lvnhmd:


packages/forms-web-app/src/controllers/register/organisation/tell-us-about-project.js
```diff
@@ -31,6 +35,21 @@ exports.postComments = async (req, res) => {
     req.session.comment = comment;
     if (mode === 'draft') {
       req.session.mode = 'draft';
+
```
cpcundill, line 38: Why did we need to update this controller @lvnhmd ?

cpcundill (changes requested):

packages/forms-web-app/src/controllers/register/organisation/tell-us-about-project.js
cpcundill: Yep - thanks for explaining 👍

cpcundill (approved):

PR #158, opened 2022-03-14
Add Service Name header to be added to each page within the registrat…

lvnhmd:
…ion journey

### DIFF
--- a/packages/forms-web-app/src/views/layouts/register-main.njk
+++ b/packages/forms-web-app/src/views/layouts/register-main.njk
@@ -1,0 +1,10 @@
+{% extends "layouts/main.njk" %}
+
+{% block header %}
+  {{ pinsHeader({
+    homepageUrl: "/",
+    serviceName: "Register to have your say about a national infrastructure project",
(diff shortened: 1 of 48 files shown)


packages/forms-web-app/src/views/layouts/register-main.njk
```diff
@@ -0,0 +1,10 @@
+{% extends "layouts/main.njk" %}
```
pk2453, line 1: Is this file needed, or any issue with main.njk?

pk2453 (commented):
I think modify serviceName in config.js and then add this line into app.js env.addGlobal('serviceName', config.serviceName); should be enough

cpcundill (approved):

packages/forms-web-app/src/views/layouts/register-main.njk
lvnhmd: @pk2453 [As per the ticket](https://pins-ds.atlassian.net/browse/ASB-162), we only need the service name in the register journey pages, main.njk is included in other pages as well

PR #175, opened 2022-03-17
ASB-124: acceptance tests added for field validation in agent journey

harish-kancharla:


pk2453 (approved):

e2e-tests/cypress/integration/registration/agent/full-name.feature
```diff
@@ -0,0 +1,40 @@
+@testSuite @registration @agent
+Feature: What is your full name page
+  As a agent registering on behalf of another person, a family or organisation I do not work for
+  I want to provide my full name
+  So that I can register on their behalf as an interested party 
+
+  Background: Start registration as an agent and progress to What is your full name page
+    Given I am registering as an "On behalf of another person, a family group or an organisation I do not work for"
```
cpcundill, line 8: I think I prefer to not parameterise this step and instead use something more succinct.  Let's add another 'Given` step with text "I am registering as an agent" or "I am registering on behalf of another party"

e2e-tests/cypress/integration/common/actions.js
```diff
@@ -1,11 +1,11 @@
 import {Given, When} from "cypress-cucumber-preprocessor/steps";
 
-Given('I am registering as an organisation', () => {
+Given('I am registering as an {string}', (radioChoice) => {
```
cpcundill, line 3: I think I prefer to not parameterise this step and instead use something more succinct. Let's add another 'Given` step with text "I am registering as an agent" or "I am registering on behalf of another party"

cpcundill (changes requested):

cpcundill (approved):

PR #182, opened 2022-03-18
ASB-170-175 modify content

pk2453:


packages/forms-web-app/src/views/register/start.njk
```diff
@@ -53,7 +53,7 @@
         <li>registration comments about the project</li>
       </ul>
       <h3 class="govuk-heading-s">If you are registering for an organisation you work or volunteer for</h3>
-      <p>You will need:</p>
+      <p>You will need your:</p>
```
cpcundill, line 56: Should this `your` have been added?  This would now read like so: `[You will need your] your full name` and `[You will need your] the name of the charity or organisation you work for`

cpcundill (changes requested):

packages/forms-web-app/src/views/register/start.njk
pk2453: This is updated, seems I did it by mistake.

PR #196, opened 2022-03-23
Render page 'missed deadline' when registration period closed

lvnhmd:


cpcundill (changes requested):
Can we also take this opportunity to delete the unused view `register/token-expired.njk`?

cpcundill (approved):

PR #199, opened 2022-03-24
Do not display over 18 for organisation

lvnhmd:


packages/forms-web-app/src/views/register/agent/cya/cya-representee-detail.njk
```diff
@@ -19,7 +19,104 @@
 {% endif %}
 
 <h2 class="govuk-heading-m">Who you are representing?</h2>
-{{ govukSummaryList({
```
cpcundill, line 22: Is there a way to build up the rows dynamically in Nunjucks?  That way we could call `govukSummaryList` just once with the right rows data.

cpcundill (changes requested):

packages/forms-web-app/src/views/register/agent/cya/cya-representee-detail.njk
```diff
@@ -39,8 +35,9 @@
                 }
               ]
             }
-          },
-          {
+          }] %}
+{% if data.representing !== 'organisation' %}
+  {% set rows = (rows.push({
```
cpcundill, line 40: rows.push can take an array of objects so no need to keep calling rows.push for each individual object after the conditional one (over 18)

cpcundill (changes requested):

cpcundill (approved):
Thanks for making the improvements 👍

PR #208, opened 2022-03-28
refactor layout

pk2453:


cpcundill (changes requested):
I think this PR is a good example why we should use JIRA tickets for each item of refactoring so that we gain a team consensus on what we want to improve and how.

The original intention behind Elvin creating a separate layout for the registration pages was because the service name for private beta version 1 is different.  However, when we launch all of the project pages, I expect the service name will change and become more generic.

Can you please take this one back to the drawing board by creating a JIRA ticket to explain the intention of the change and then we can discuss at refinement.

PR #219, opened 2022-04-05
Trigger automatic pre-commit

lvnhmd:


cpcundill (changes requested):
Looking good - can we please also add a check in the build in case committers haven't installed the pre-commit checks?   You can do this by simply adding another `npm run` command within the Dockerfiles

packages/applications-service-api/Dockerfile
```diff
@@ -12,15 +12,16 @@ ADD api api
...
 ADD package.json package.json
 ADD .sequelizerc .sequelizerc
 ADD package-lock.json package-lock.json
 
-RUN npm i
-RUN npm run documentation
-RUN npm prune --production
+RUN npm i --production
+RUN npm run lint
+RUN npm run documentation; echo $? > /npm.exitcode
```
cpcundill, line 24: This looks like a merge problem @lvnhmd  - I removed the `npm.exitcode` approach from the Dockerfiles

cpcundill (changes requested):

packages/applications-service-api/Dockerfile
lvnhmd: @cpcundill - Yes, I had a conflict (which I was afraid of), I seem to have missed this, what is the change you are requesting if you have already fixed it ?

cpcundill (changes requested):
I think we need to update the `Dockerfile`s within the `pipelines` directory too

packages/applications-service-api/Dockerfile
```diff
@@ -12,15 +12,16 @@ ADD api api
...
 ADD jest.config.js jest.config.js
+ADD .eslintrc .eslintrc
 ADD data data
 
 ADD package.json package.json
 ADD .sequelizerc .sequelizerc
 ADD package-lock.json package-lock.json
 
-RUN npm i
+RUN npm i --production
```
cpcundill, line 22: I think this line should be just `npm i` since we need dev dependencies for the `npm run documetation` step

pipelines/applications-service-api/Dockerfile
```diff
@@ -15,8 +15,10 @@ ADD data data
 ADD package.json package.json
 ADD .sequelizerc .sequelizerc
 ADD package-lock.json package-lock.json
+ADD .eslintrc .eslintrc
 
 RUN npm i
```
lvnhmd, line 20: @cpcundill - not sure how is this happening, but this line is `RUN npm i` only ?

packages/applications-service-api/Dockerfile
lvnhmd: done

cpcundill (approved):

cpcundill (dismissed):

PR #226, opened 2022-04-13
ASB-14: pagination acceptance tests on application documents page

harish-kancharla:


e2e-tests/cypress/integration/project-application-documents.feature
```diff
@@ -0,0 +1,78 @@
...
+    Scenario: verify pagination functionality on Project application documents page
+        Then I verify below pagination is present on the page
+            | Data              |
+            | 1                 |
+            | 2                 |
+            | 3                 |
+            | Next set of pages |
+        Then I verify text "Showing 1 to 20 of 46 results" is present on the page
+        Then I verify that only "20" results present on each page
+        And I click on pagination "2"
```
cpcundill, line 20: Improved wording would be `When I navigate to page "2" of the results`

e2e-tests/cypress/integration/project-application-documents.feature
```diff
@@ -0,0 +1,78 @@
+@testSuite
+Feature: Project Application documents page
```
cpcundill, line 2: So I feel this is a good start but we don't have enough data with just this one project to really test the full extent of behaviour with the pagination controls.  We don't really explore what the conditional display of the ellipsis for example, or whether the first and last page links are always visible.

cpcundill (changes requested):
Let's add a larger data set of documents for another project so we can test for some other behaviour of the pagination controls.

cpcundill (approved):

PR #231, opened 2022-04-20
ASB-266 and ASB-271

lvnhmd:
ASB-266 API: Remove stage filter from documents endpoint
ASB-271 API: Add sorting by publication date

### DIFF
--- a/packages/applications-service-api/src/controllers/documents.js
+++ b/packages/applications-service-api/src/controllers/documents.js
@@ -29,14 +29,13 @@ module.exports = {
       const { itemsPerPage, documentsHost } = config;
       const totalItems = documents.count;
       const byType = R.groupBy(R.prop('type'));
-      const byStage = R.groupBy(R.prop('Stage'));
       const rows = documents.rows.map((row) => ({
         ...row.dataValues,
         path: row.dataValues.path ? `${documentsHost}${row.dataValues.path}` : null,
       }));
 
       const wrapper = {
-        documents: Object.entries(byStage(rows)).map((e) => ({ [e[0]]: byType(e[1]) })),
+        documents: byType(rows),
         totalItems,
         itemsPerPage,
         totalPages: Math.ceil(totalItems / itemsPerPage),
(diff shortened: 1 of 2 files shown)


packages/applications-service-api/src/controllers/documents.js
```diff
@@ -29,14 +29,13 @@ module.exports = {
...
       const byType = R.groupBy(R.prop('type'));
-      const byStage = R.groupBy(R.prop('Stage'));
       const rows = documents.rows.map((row) => ({
         ...row.dataValues,
         path: row.dataValues.path ? `${documentsHost}${row.dataValues.path}` : null,
       }));
 
       const wrapper = {
-        documents: Object.entries(byStage(rows)).map((e) => ({ [e[0]]: byType(e[1]) })),
+        documents: byType(rows),
```
cpcundill, line 38: We will actually need documents to be returned in a different structure and not to be grouped.

cpcundill (changes requested):

cpcundill (approved):

packages/applications-service-api/src/routes/documents.js
```diff
@@ -5,5 +5,6 @@ const documentsController = require('../controllers/documents');
 const router = express.Router();
 
 router.post('/:caseRef', documentsController.getDocuments);
+router.post('/', documentsController.getV2Documents);
```
cpcundill, line 8: Should be a `router.get`

cpcundill (changes requested):

packages/applications-service-api/src/controllers/documents.js
```diff
@@ -55,4 +55,53 @@ module.exports = {
...
+    const { caseRef } = req.query;
+    const { pageNo = 1, searchTerm = null, filters = null } = req.body;
+
+    if (!caseRef) {
+      throw ApiError.badRequest('Required query parameter caseRef missing');
+    }
+
+    logger.debug(`Retrieving documents by case reference ${caseRef} ...`);
+    try {
+      const documents = await getDocumentsFromDocumentsApiService(
```
cpcundill, line 69: Looks like we'll also need to call a different business service method since the current one still includes the filter for only stages 1, 2 and 3.  You need to only add the `order: [['date_published', 'DESC']],` within this new business service method too in order to preserve v1 functionality untouched.

cpcundill (changes requested):

cpcundill (changes requested):
Looking good - will there by some unit tests added?

cpcundill (approved):

PR #233, opened 2022-04-22
ASB-269 change layout and styling

pk2453:


pk2453:
Slight change required in internal layout of about-your-application page which can be done during changing the layout of documents displayed. Other pages header and banner is also changed.

cpcundill:
Checking out locally to take a look at the visual changes

cpcundill (changes requested):
Have scheduled a meeting at 3pm with Naomi and Aqib to confirm whether we should be going with the bigger width.

cpcundill (changes requested):
Still not quite what I'd expected and tried to explain earlier.  I've mocked up some changes locally for us to discuss on Monday.

cpcundill:
Will be raising a new approach with a revised approach

PR #236, opened 2022-04-25
Remove unused build pipelines and update Dockerfiles

andylangridge:
- Deletes old build pipelines
- Changes build context of Dockerfiles to the root of the project - This allows copying the common package directly into the forms-web-app image instead of relying on a Docker image that has been pushed.
- Dockerfiles now multi-stage to reduce final image size


cpcundill (dismissed):
Looks good to me 👍

packages/applications-service-api/Dockerfile
```diff
@@ -14,18 +14,23 @@ RUN npm i
...
 
 COPY packages/applications-service-api/__tests__ ./__tests__
 COPY packages/applications-service-api/jest.config.js jest.config.js
-COPY packages/applications-service-api/.eslintrc .eslintrc
+COPY packages/applications-service-api/.eslint* ./
+COPY packages/applications-service-api/.prettier* ./
 
 RUN npm run lint
-RUN npm run documentation
+RUN npm run documentation; echo $? > /npm.exitcode
```
cpcundill, line 27: Doesn't this swallow the error?  We want test failures to fail the build

packages/forms-web-app/Dockerfile
```diff
@@ -15,19 +15,24 @@ RUN npm run build
...
 COPY --from=build /opt/common ../common
 
 COPY packages/forms-web-app/__tests__ ./__tests__
 COPY packages/forms-web-app/jest.config.js jest.config.js
-COPY packages/forms-web-app/.eslintrc .eslintrc
+COPY packages/forms-web-app/.eslint* ./
+COPY packages/forms-web-app/.prettier* ./
 
 RUN npm run lint
 RUN npm run documentation; echo $? > /npm.exitcode
```
cpcundill, line 29: Same here - doesn't this swallow the error?  We want test failures to fail the build

cpcundill (changes requested):

packages/applications-service-api/Dockerfile
andylangridge: Yeah it does, but the pipeline contains a step that publishes the results to Azure DevOps "Test Plans" section. This publishes the results, and fails the build if any test failures.

If this build step fails, then the report step never publishes its results. Hence we swallow the error here :)

cpcundill (approved):

PR #238, opened 2022-04-25
Implement search box and search submission

pk2453:


packages/forms-web-app/src/routes/projects/documents.js
```diff
@@ -6,7 +6,7 @@ const aboutTheApplicationController = require('../../controllers/projects/docume
 
 router.get('/:case_ref', projectsController.getExamination);
 router.get('/:case_ref/documents/:page', aboutTheApplicationController.getAboutTheApplication);
-router.post('/:case_ref/documents/search/:page', aboutTheApplicationController.postSearchDocument);
-router.post('/:case_ref/documents/filter/:page', aboutTheApplicationController.postFilterDocument);
+router.post('/:case_ref/documents/:page', aboutTheApplicationController.postSearchDocument);
```
cpcundill, line 9: We need the route to be navigable and shareable so let's just use the existing `get` route and pass in a query parameter for the search term

packages/forms-web-app/src/views/projects/documents.njk
```diff
@@ -1,26 +1,30 @@
 {% extends "layouts/project-3-columns.njk" %}
 
 {% set activeProjectLink = 'about-the-application' %}
-
-{% block columnTwo %}
+{% block resultSec %}
   <div class="govuk-grid-row">
     <h2 class="govuk-heading-l">Project application documents</h2>
-
+      <form action="/projects/{{ caseRef }}/documents/1" method="POST" novalidate>
```
cpcundill, line 7: As per my route comment - this doesn't need to set the `method` attribute since we will use the default `GET` method.

cpcundill (changes requested):
A few comments to address and a catchup merge needed

packages/forms-web-app/src/routes/projects/documents.js
pk2453: Will do overall separate implementation once Elvin's code is merged, Would be easier to do all straight away. Its POST based on current implementation and can be changed to GET once Elvin's code is ready.

PR #244, opened 2022-04-28
Feature/asb 279 filter options and counts

lvnhmd:


packages/applications-service-api/src/services/document.service.js
```diff
@@ -47,11 +47,56 @@ const getDocuments = async (caseRef, pageNo, searchTerm) => {
...
+    }
+
+    if (type) {
+      where[Op.and].push({
+        filter_1: { [Op.in]: type },
+      });
+    }
+
+    where[Op.and].push({
+      [Op.or]: orOptions,
```
cpcundill, line 97: Would this fail when `orOptions` is not defined?

packages/applications-service-api/src/services/document.service.js
```diff
@@ -47,11 +47,56 @@ const getDocuments = async (caseRef, pageNo, searchTerm) => {
...
 };
 
-const getOrderedDocuments = async (caseRef, pageNo) => {
+const getOrderedDocuments = async (caseRef, pageNo, searchTerm, stage, type) => {
   const { itemsPerPage: limit } = config;
   const offset = (pageNo - 1) * limit;
 
-  const where = { case_reference: caseRef };
+  let where = { case_reference: caseRef };
+  if (stage) where = { ...where, Stage: { [Op.in]: stage } };
```
cpcundill, line 55: What's the difference between this where condition and that on line 84?

packages/applications-service-api/src/services/document.service.js
```diff
@@ -47,11 +47,56 @@ const getDocuments = async (caseRef, pageNo, searchTerm) => {
...
+  if (type) where = { ...where, filter_1: { [Op.in]: type } };
```
cpcundill, line 56: What's the difference between this where condition and that on line 90?

packages/applications-service-api/src/services/document.service.js
```diff
@@ -47,11 +47,56 @@ const getDocuments = async (caseRef, pageNo, searchTerm) => {
   return documents;
 };
 
-const getOrderedDocuments = async (caseRef, pageNo) => {
+const getOrderedDocuments = async (caseRef, pageNo, searchTerm, stage, type) => {
   const { itemsPerPage: limit } = config;
   const offset = (pageNo - 1) * limit;
 
-  const where = { case_reference: caseRef };
+  let where = { case_reference: caseRef };
```
cpcundill, line 54: We need to make sure the base where condition includes `Stage: { [Op.gt]: 0 }`

See https://github.com/Planning-Inspectorate/applications-service/pull/237/files#diff-b4447c6a47b8ab588745de2cd533acd987452209f382cc2fe2315fde320f53f0R54

cpcundill (changes requested):

cpcundill (changes requested):
Did I miss something or shouldn't there be queries to get distinct lists of stages and types (along with counts for each) from the documents that belong to a given project?

packages/applications-service-api/src/services/document.service.js
cpcundill: ...but when we don't have a searchTerm?

lvnhmd: When we do not have searchTerm, we do not declare and use orOptions at all : 
```
if (searchTerm) {
    const orOptions = [
      { ...
```

packages/applications-service-api/src/services/document.service.js
lvnhmd: On line 84 we create AND construct - for the orOptions AND the rest. Line 55 translates to `{ case_reference: 'TR040009', Stage: { [Symbol(in)]: [ '0', '1' ] } } `, whilst line 84 translates to `{ [Symbol(and)]: [ { case_reference: 'TR040009' }, { Stage: [Object] } ] }`

packages/applications-service-api/src/services/document.service.js
lvnhmd: On line 90 we use AND construct - for the orOptions AND the rest. Line 55 translates to `{ case_reference: 'TR040009', filter_1: { [Symbol(in)]: [ 'f1', 'f2' ] } } `, whilst line 84 translates to `{ [Symbol(and)]: [ { case_reference: 'TR040009' }, { filter_1: [Object] } ] }`

packages/applications-service-api/src/services/document.service.js
lvnhmd: Can we ever have a Stage value less than 0 ?

cpcundill: Not by intention from the UI but we need to make sure we always filter out documents with Stage 0 since they are the documents attached to registration comments!

packages/applications-service-api/src/services/document.service.js
```diff
@@ -63,7 +108,17 @@ const getOrderedDocuments = async (caseRef, pageNo) => {
   return documents;
 };
 
+const getFilters = async (filter) => {
```
cpcundill, line 111: Appreciate you wanted to keep this generic but we do need to exclude Stage 0  for the stage filter

cpcundill (changes requested):

packages/applications-service-api/src/services/document.service.js
lvnhmd: Done.

packages/applications-service-api/src/services/document.service.js
```diff
@@ -47,11 +47,56 @@ const getDocuments = async (caseRef, pageNo, searchTerm) => {
   return documents;
 };
 
-const getOrderedDocuments = async (caseRef, pageNo) => {
+const getOrderedDocuments = async (caseRef, pageNo, searchTerm, stage, type) => {
   const { itemsPerPage: limit } = config;
   const offset = (pageNo - 1) * limit;
 
-  const where = { case_reference: caseRef, Stage: { [Op.gt]: 0 } };
+  let where = { case_reference: caseRef };
```
cpcundill, line 54: We need the exclusion of Stage 0 to be re-instated: `, Stage: { [Op.gt]: 0 }`

cpcundill (changes requested):

cpcundill (approved):

PR #249, opened 2022-05-05
Feature/asb 296 api new representations endpoint

lvnhmd:


cpcundill (approved):

packages/applications-service-api/src/services/representation.service.js
```diff
@@ -0,0 +1,9 @@
+const db = require('../models');
+
+const getRepresentationsForApplication = async (applicationId) => {
+  return db.Representation.findAll({ where: { CaseReference: applicationId } });
```
cpcundill, line 4: We need to paginate the response in a similar way to documents

cpcundill (changes requested):

cpcundill (approved):

PR #252, opened 2022-05-09
ASB-281 add filters

pk2453:


cpcundill (changes requested):

cpcundill (changes requested):
As discussed, happy that the sorting and slicing of the Type filters is done on the frontend for now.  However, we need to sort the top 5 by type name ASC after the slicing.

packages/forms-web-app/src/controllers/projects/documents.js
```diff
@@ -1,26 +1,57 @@
...
       queryUrl = `?searchTerm=${params.searchTerm}`;
     }
     const respData = response.data;
-    const { documents } = respData;
+    const { documents, filters } = respData;
+    const { stageFilters, typeFilters } = filters;
     logger.debug(`Document data received:  ${JSON.stringify(documents)} `);
     const paginationData = getPaginationData(respData);
     const pageOptions = calculatePageOptions(paginationData);
+    var customStageFilters = [];
```
cpcundill, line 27: Why the name `customStageFilter`?  What's custom about them?

cpcundill (dismissed):

cpcundill (approved):

PR #253, opened 2022-05-09
add projectName from API

pk2453:


cpcundill (changes requested):
Can we update the unit test to check the expected projectName is passed for rendering?  We'll need a mock of the call for project data.

pk2453:
> Can we update the unit test to check the expected projectName is passed for rendering? We'll need a mock of the call for project data.

Test added

PR #256, opened 2022-05-09
Feature/asb 305 filter options and counts

lvnhmd:


packages/applications-service-api/src/services/representation.service.js
```diff
@@ -1,20 +1,65 @@
+const { Op } = require('sequelize');
 const db = require('../models');
 const config = require('../lib/config');
 
-const getRepresentationsForApplication = async (applicationId, page) => {
+const getRepresentationsForApplication = async (applicationId, page, searchTerm) => {
   const { itemsPerPage: limit } = config;
   const offset = (page - 1) * limit;
-  const where = { CaseReference: applicationId };
+  let where = { CaseReference: applicationId };
```
cpcundill, line 8: I think it's better to set this to:

```
let where = { [Op.and]: [{ CaseReference: applicationId }] };
```

Then you can remove the re-assignment on line 29.

cpcundill (changes requested):

cpcundill (approved):

PR #257, opened 2022-05-09
Feature/asb 304 extend representations endpoint with filter parameters

lvnhmd:


packages/applications-service-api/src/controllers/representations.js
```diff
@@ -1,23 +1,37 @@
...
+} = require('../services/representation.service');
 
 const ApiError = require('../error/apiError');
 
 module.exports = {
   async getRepresentationsForApplication(req, res) {
-    const { applicationId, page } = req.query;
+    const { applicationId, page, searchTerm, type } = req.query;
+
+    const slugified = type && !(type instanceof Array) ? [type] : type;
```
cpcundill, line 18: The term "slugified" refers to a type of string casing where spaces are replaced with hyphens so they can be better used in URLs.  It doesn't work well here.  To be honest I think we should always treat the `type` filter as an array like so:
```
const types = type instanceof Array ? [...type] : [type];
```

packages/applications-service-api/src/services/representation.service.js
```diff
@@ -1,20 +1,72 @@
...
 const db = require('../models');
 const config = require('../lib/config');
 
-const getRepresentationsForApplication = async (applicationId, page) => {
+const getRepresentationsForApplication = async (applicationId, page, searchTerm, type) => {
   const { itemsPerPage: limit } = config;
   const offset = (page - 1) * limit;
-  const where = { CaseReference: applicationId };
+  let where = { CaseReference: applicationId };
+  if (type) where = { ...where, RepFrom: { [Op.in]: type } };
```
cpcundill, line 9: We should just keep one form of query object - i.e. the form with the [Op.and] key at the top-level.   Then this line could become:

```
if (type) {
      where[Op.and].push({
        RepFrom: { [Op.in]: type },
      });
    }
```

The code between lines 32 and 36 can then also be removed.

cpcundill (changes requested):

cpcundill (approved):

PR #259, opened 2022-05-10
ASB-184: acceptance tests added for search functionality on project d…

harish-kancharla:
…ocuments page

### DIFF
--- a/e2e-tests/cypress/integration/project-application-documents.feature
+++ b/e2e-tests/cypress/integration/project-application-documents.feature
@@ -186,3 +186,41 @@ Feature: Project Application documents page
         Given I navigate to "Ho Ho Hooo" project Overview page
         When I click on "Project application documents" link
         Then I verify that no documents found text displayed on the page
+
+    Scenario: search returns matching documents
+        Given I navigate to "Hinkley Point C New Nuclear Power Station Material Change 1" project Overview page
+        When I click on "Project application documents" link
+        When I enter text "consultee" into search field
+        And I click on search button
+        Then I can verify that below project documents were returned
+            | Document                                                                                                                                                                                                                                                                                                                                                                                                                                                                |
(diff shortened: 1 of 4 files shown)


e2e-tests/cypress/integration/project-application-documents.feature
```diff
@@ -186,3 +186,41 @@ Feature: Project Application documents page
...
+        When I click on "Project application documents" link
+        When I enter text "consultee" into search field
+        And I click on search button
+        Then I can verify that below project documents were returned
+            | Document                                                                                                                                                                                                                                                                                                                                                                                                                                                                |
+            | \n            \n              Consultee List sent to the Secretary of State (, 0Bytes)\n            \n            \n              \n                Published by EDF Energy\n              \n            \n            \n              \n                2019-04-16\n              \n              \n                Pre Application\n              \n              \n                Material Change\n              \n            \n                                   |
```
cpcundill, line 197: I think we can do better than this by stripping out the whitespace/tabs from the content on screen

e2e-tests/cypress/integration/project-application-documents.feature
```diff
@@ -186,3 +186,41 @@ Feature: Project Application documents page
         Given I navigate to "Ho Ho Hooo" project Overview page
         When I click on "Project application documents" link
         Then I verify that no documents found text displayed on the page
+
+    Scenario: search returns matching documents
+        Given I navigate to "Hinkley Point C New Nuclear Power Station Material Change 1" project Overview page
```
cpcundill, line 191: Would be better to use the Cleve Hill Solar Farm project which was added specifically because it has an export of production document data so is more realistic.  For one thing it has proper file size values 😁

cpcundill (changes requested):

e2e-tests/cypress/integration/project-application-documents.feature
harish-kancharla: changed project as suggested

e2e-tests/cypress/integration/project-application-documents.feature
harish-kancharla: stripped whitespace and line breaks

cpcundill (approved):

PR #260, opened 2022-05-10
add everythingelse type filter

pk2453:


packages/forms-web-app/src/controllers/projects/documents.js
```diff
@@ -55,15 +68,13 @@ function renderData(req, res, searchTerm, params, response, projectName, stageLi
           checked: typeList.includes(type.name),
         });
       }, Object.create(null));
-
-    typeFilters.slice(-(typeFilters.length - 5)).forEach(function (type) {
-      otherTypeFilters.push({
-        text: `${type.name} (${type.count})`,
-        value: type.name,
-        checked: typeList.includes(type.name),
+    if (top5TypeFilters.length > 5) {
```
cpcundill, line 71: We need to change some variable names since this is now looking very confusing.  Top 5 filters being greater than 5? 🤨   Isn't the backend now returning up to the top 5 plus the everything else filter?

cpcundill (changes requested):

packages/forms-web-app/src/controllers/projects/documents.js
pk2453: sorry this meant to be typeFilters so updated now.

cpcundill (approved):

PR #261, opened 2022-05-12
Feature/asb 310 endpoint to fetch a single representation

lvnhmd:


cpcundill (changes requested):
Needs a catchup with `main` branch

packages/applications-service-api/__tests__/unit/controllers/representations.test.js
```diff
@@ -70,42 +78,37 @@ const returnData = {
...
-});
-
 jest.mock('../../../src/lib/config.js', () => ({
   logger: {
     level: process.env.LOGGER_LEVEL || 'info',
   },
   itemsPerPage: 3,
 }));
 
+jest.mock('../../../src/services/representation.service');
```
cpcundill, line 91: I appreciate it is much easier to mock the servicer due to the complexity of mocking Sequelize.  However, we are now missing a test coverage for the representation business service.  Happy for that to be a separate unit test file, but I think we shouldn't avoid it due to the difficulty.  Let's just find a solution together to manage the complexity.

cpcundill (changes requested):

packages/applications-service-api/src/services/representation.service.js
```diff
@@ -43,6 +47,27 @@ const getRepresentationsForApplication = async (applicationId, page, searchTerm)
   return representations;
 };
 
+const getRepresentationById = async (ID) => {
+  return db.Representation.findOne({ where: { ID } });
+};
+const getFilters = async (filter, applicationId) => {
```
cpcundill, line 53: This shouldn't be needed right?

cpcundill (changes requested):

packages/applications-service-api/__tests__/unit/controllers/representations.test.js
cpcundill: What do you think @lvnhmd ?

cpcundill (changes requested):

packages/applications-service-api/__tests__/unit/controllers/representations.test.js
lvnhmd: @cpcundill - I am working on it. Opted for a separate unit test file.

cpcundill (approved):

PR #263, opened 2022-05-13
ASB-185: Handle no documents available vs no documents found properly

cpcundill:
Changed v2 documents endpoint to longer return 404 but instead an empty list when no documents are found.   Changed frontend to handle an empty list of documents with no search and filters differently than when there is an empty list of documents due to search and/or filtering


cpcundill:
Superseded by https://github.com/Planning-Inspectorate/applications-service/pull/264

PR #267, opened 2022-05-13
ASB-300 add search box for representation

pk2453:


packages/forms-web-app/src/controllers/projects/representations.js
```diff
@@ -18,6 +21,7 @@ exports.getRepresentations = async (req, res) => {
       representations: representationsResponse.data.representations,
       paginationData,
       pageOptions,
+      searchTerm,
```
cpcundill, line 24: Need a `queryUrl` containing the searchTerm to be passed down for use in the pagination bar.  See documents as a reference example.

cpcundill (changes requested):

packages/forms-web-app/src/controllers/projects/representations.js
pk2453: It was already added, jest need to refresh as it was in 2nd commit

cpcundill: I think you're missing the corresponding `{{queryUrl}}` being appended to the URL used within the pagination bar, i.e. `pageLinkTemplate`?

cpcundill (changes requested):

packages/forms-web-app/src/controllers/projects/representations.js
pk2453: added to pagination now

cpcundill (approved):

PR #296, opened 2022-06-20
Run Linting and tests from repo root

andylangridge:
## Summary
- Updated prettier config and make it run from project root correctly
- Updated ESLint config, can also be run from root
- Updated Jest config, can also be run from root
- Install Husky and make hooks required - these now run linting and tests
- Run ESLint and Prettier against entire repo
- Removed a bunch of unused dependencies from the package.json files
- Removed the "pre-commit" config file, since we are using Husky and don't need 2 tools for git hooks.
- Fixed Commitizen config

### Jest
I have updated the config to stop running coverage reports on every execution, and instead have made this an option via the npm scripts. I'm also now generating the coverage report in the format that Azure DevOps requires and am saving all the reports Jest generates in a "jest-reports" folder. There is now a shared jest config file in the root, and each package uses this shared config as well as any other package specific options.

In order to run the tests from the root, I used `npm-run-all` to handles this.

### ESLint
Updated the config so there is a shared file in the root directory. It appears that all the code, even in the frontend, is to be run by Node instead of the browser... Except the public/javascripts folder, where I added another eslint config file to accout for this. It would be nice if ESLint didn't need to be installed a dependency anywhere but the root package.json, but since each package might have it's own config, I couldn't figure out how to get round that.

Similar to the tests, using `npm-run-all` to run linting in all packages.

### Prettier
Disabled ESLint formatting rules so that Prettier now handles all of this. To do this I installed the `eslint-config-prettier` package.

### Unused dependencies
Since Applications was copied from Appeals, there were a load of packages being installed that weren't being used. I just ran an `npx depcheck` to find them. Have removed the offending packages, but I did double-check before deleting since some were being used with the tool unable to tell. So please check I've not removed anything vital!

### Commitizen
Ran `npx commitizen init cz-conventional-changelog --save-dev` to fix the `npm run commit` script. It should now run a CLI tool to guide you through the commit format that is expected. One of the Husky hooks ensures that the commit needs to be in the correct format before it is accepted.


.prettierignore
```diff
@@ -0,0 +1,4 @@
+jest-reports
+README.md
+*.njk
```
Philip-Philippou1, line 3: Perhaps one for a future PR but we may be able to add.njk files as an override.  Something like:
https://github.com/prettier/prettier-vscode/issues/1405

Philip-Philippou1 (approved):

.prettierignore
andylangridge: Unfortunately, this condensed them into 1 line and there are also some errors. So definitely one for a future PR me thinks!

PR #301, opened 2022-06-22
fix: documentHost prefix added to reg. comment

ozankc:


packages/applications-service-api/src/controllers/representations.js
```diff
@@ -72,10 +72,26 @@ module.exports = {
...
-			representation.dataValues.attachments = Object.values(attachments);
+			let attachments = await getDocumentsByDataId(dataIDs);
 
+			if (
+				attachments &&
+				attachments['0'] &&
+				Object.keys(attachments['0']).length === 0 &&
+				Object.getPrototypeOf(attachments['0']) === Object.prototype
+			) {
+				console.log('has no attachment');
```
Philip-Philippou1, line 86: Can we use the logger here?

ozankc: logger added, thanx

Philip-Philippou1 (approved):

PR #307, opened 2022-06-27
ASB-411: Hide - Save & Exit Journey

christopher-semmens:
### Note:
WIP: Environment variables need to be configured correctly. Environment variables are working as needed in the Nunjucks views but we need to disable the logic which runs for the save and exit journey. Also note that we may need to configure the Azure environment with the new feature flag environment variables.

### Task
https://pins-ds.atlassian.net/jira/software/c/projects/ASB/boards/166?modal=detail&selectedIssue=ASB-411

### Description
- Add feature flag used in registration journey 'tell-us-about-project' views and controllers


Philip-Philippou1 (commented):
This PR breaks the e2tests, because the button no longer exists. Example feature file is :
e2e-tests/cypress/integration/registration/organisation/what-do-you-want-to-tell-about-project.feature

Test - 1) User saves with no comment 
Few options here,
1) We may be able to activate the button in the local env by setting feature in docker-compose.yml as I think that is only for local dev
2) Deactivate these tests until functionality is worked on.
3) Somehow use feature flag in the tests.

There may be other ways.

Philip-Philippou1 (approved):
LGTM

Philip-Philippou1 (approved):

PR #316, opened 2022-07-06
docs(contact us for support): content change for contact us (previous…

Xfarya:
This one should be working - it's simple content change to contact us for support (pro talk to an advisor)

### DIFF
--- a/packages/forms-web-app/src/views/projects/examination.njk
+++ b/packages/forms-web-app/src/views/projects/examination.njk
@@ -66,6 +66,26 @@
     govuk-section-break--xl
     govuk-section-break--visible"/>
 
+ <h2 class="govuk-heading-m">Contact us for support</h2>
+          <p class="govuk-body"> 
+            <strong>
+              Project case team email:
+            </strong>
+             </br>
+                <a class="govuk-link" href="mailto:{{ appData.ProjectEmailAddress }}">{{ appData.ProjectEmailAddress }}
+                </a>
+            </p>
+           <p class="govuk-body"> 
+          <strong>
+            General enquiries telephone:
+          </strong></br>
+          <a href="tel:03034445000">0303 444 5000</a></br>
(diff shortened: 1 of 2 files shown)


packages/forms-web-app/src/views/projects/examination.njk
```diff
@@ -66,6 +66,26 @@
     govuk-section-break--xl
     govuk-section-break--visible"/>
 
+ <h2 class="govuk-heading-m">Contact us for support</h2>
+          <p class="govuk-body">
```
JamesP-KC, line 70: Looks like this `<p>` is indented a bit much?

packages/forms-web-app/src/views/projects/examination.njk
```diff
@@ -66,6 +66,26 @@
     govuk-section-break--xl
     govuk-section-break--visible"/>
 
+ <h2 class="govuk-heading-m">Contact us for support</h2>
+          <p class="govuk-body"> 
+            <strong>
+              Project case team email:
+            </strong>
+             </br>
+                <a class="govuk-link" href="mailto:{{ appData.ProjectEmailAddress }}">{{ appData.ProjectEmailAddress }}
```
JamesP-KC, line 75: This `<a>` tag doesn't need to be indented (`</br>` is self-closing so the `<a>` isn't nested within it)

JamesP-KC (commented):
Just some minor formatting issues. Your editor might have a way of auto-formatting the file which might be worth enabling if available.

packages/forms-web-app/src/views/projects/examination.njk
Xfarya: Hi James, thank you - you're right about formatting, I do have prettier in my docs (I am using VSCode), so I thought it would auto-format once I saved the file?

Philip-Philippou1 (dismissed):

PR #334, opened 2022-07-13
feat(Hide project timeline link from project overview page): hide Pro…

mauriciokin:
JIRA TICKET: https://pins-ds.atlassian.net/jira/software/c/projects/ASB/boards/166?modal=detail&selectedIssue=ASB-334

Hide Project Timeline Link from Project Overview Page.

Commented out the pertinent parts of the code where we want the element "Project Timeline" hidden.

Got it commented out and with this JIRA ticket reference so it's easy to make it get displayed back when it's time.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -68,7 +68,7 @@ services:
       FEATURE_REDIS_SESSION_STORE: 'true'
       FEATURE_SAVE_AND_EXIT_OPTION: 'false'
       FEATURE_SHOW_AFFECTED_AREA_SECTION: 'false'
-      FEATURE_HIDE_ROUTE: 'true'
+      FEATURE_PROJECT_TIMELINE_LINK: 'false'
     ports:
       - '9004:3000'
       # node debugger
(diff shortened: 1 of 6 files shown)


docker-compose.yml
```diff
@@ -68,6 +68,7 @@ services:
       FEATURE_REDIS_SESSION_STORE: 'true'
       FEATURE_SAVE_AND_EXIT_OPTION: 'false'
       FEATURE_SHOW_AFFECTED_AREA_SECTION: 'false'
+      FEATURE_HIDE_ROUTE: 'true'
```
Philip-Philippou1, line 71: Can this be a specific feature name that is activated or inactive. Something like  
FEATURE_PROJECT_TIMELINE_LINK = 'false'

is perhaps more readable and consistent with other feature flags.

mauriciokin: Makes sense. I've tried to add a broader tag for possible future hiden routes too...

JamesP-KC (approved):

PR #338, opened 2022-07-15
Feature/asb330 registration comments

Xfarya:


packages/forms-web-app/src/views/layouts/vertical-tabs.njk
```diff
@@ -23,8 +23,9 @@
     </li>
     <li class="side-menu govuk-body-s {% if 'representations' == activeProjectLink %} active {% endif %}">
       {% if 'representations' != activeProjectLink %}
-        <a href="/projects/{{ caseRef }}/representations">Registration comments</a>
-      {% else %} Registration comments {% endif %}
+        <a href="/projects/{{ caseRef }}/representations">Registration comments </br> (Relevant representations)
```
Philip-Philippou1, line 26: Do we need the line break </br>

packages/forms-web-app/src/views/layouts/vertical-tabs.njk
```diff
@@ -23,8 +23,9 @@
...
+        </a>
+      {% else %} Registration comments </br> (Relevant representations) {% endif %}
```
Philip-Philippou1, line 28: Same, here is the </br> intentional?

Xfarya: Yes, I did it intentionally, only because in the prototype it is on separate lines (I like the way it looks better) -- what do you think?

Philip-Philippou1 (approved):

PR #342, opened 2022-07-19
feat(add autocomplete to register inputs): add autocomplete attr

mauriciokin:
JIRA Ticket: https://pins-ds.atlassian.net/jira/software/c/projects/ASB/boards/166?modal=detail&selectedIssue=ASB-341

Add customised autocomplete attribute to all register (govukInput) inputs

MDN Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

### DIFF
--- a/packages/forms-web-app/src/views/register/agent/name-of-organisation.njk
+++ b/packages/forms-web-app/src/views/register/agent/name-of-organisation.njk
@@ -37,6 +37,7 @@
             },
             id: "organisation-name",
             name: "organisation-name",
+						autocomplete: 'organization',
             value: organisationName or errors['organisation-name'].value,
             attributes: {"data-cy": "organisation-name"},
             errorMessage: errors['organisation-name'] and {
--- a/packages/forms-web-app/src/views/register/organisation/name-of-organisation-or-charity.njk
+++ b/packages/forms-web-app/src/views/register/organisation/name-of-organisation-or-charity.njk
@@ -37,6 +37,7 @@
             },
             id: "organisation-name",
             name: "organisation-name",
+						autocomplete: 'organization',
             value: organisationName or errors['organisation-name'].value,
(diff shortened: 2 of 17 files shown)


packages/forms-web-app/src/views/register/organisation/what-job-title-or-role.njk
```diff
@@ -34,6 +34,7 @@
             },
             id: "role",
             name: "role",
+						autocomplete: 'organization-title',
```
Philip-Philippou1, line 37: Not big at all but slightly misleading name as this is the person's role in the organisation - organisation-role

packages/forms-web-app/src/views/register/organisation/name-of-organisation-or-charity.njk
```diff
@@ -37,6 +37,7 @@
             },
             id: "organisation-name",
             name: "organisation-name",
+						autocomplete: 'organization',
```
Philip-Philippou1, line 40: organi**s**ation is used throughout the code so we should probably keep consistent.

packages/forms-web-app/src/views/register/agent/name-of-organisation.njk
```diff
@@ -37,6 +37,7 @@
             },
             id: "organisation-name",
             name: "organisation-name",
+						autocomplete: 'organization',
```
Philip-Philippou1, line 40: organisation to make it consistent with rest of the code.

JamesP-KC (approved):

packages/forms-web-app/src/views/register/organisation/what-job-title-or-role.njk
mauriciokin: I noticed that the issue is: MDN docs don't have organisation with S in their docs as reference. Only with Z

PR #355, opened 2022-07-21
Fix: Increase in character length on form submission

christopher-semmens:
### Task
https://pins-ds.atlassian.net/jira/software/c/projects/ASB/boards/166?modal=detail&selectedIssue=ASB-481

### Description

- Decode/Encode/Sanitise String functions refactored to CommonJS modules and test created for each.
- sanitise-string function decodes string before stripping tags. This prevents duplicate string encoding.
- The comments data is decoded in Express middleware before validation. We now validate against the length of characters of the users input (minus tags) and not the encoded length. This will prevent the user  receiving an inaccurate error when their input has not exceeded the max length but exceeds it when encoded
- sanitise-form.js refactored to utilise CommonJS and compiled by Webpack. Improved conditionals to make the code more robust
- Test created for sanitise-form.js that checks the fetch payload


mauriciokin (approved):

mauriciokin (approved):
legit

PR #356, opened 2022-07-22
Fix: Sanitise Form 429 Error

christopher-semmens:


packages/forms-web-app/src/lib/decode-string.js
```diff
@@ -1,7 +1,13 @@
 const decodeString = (stringToDecode) => {
 	if (!stringToDecode || typeof stringToDecode !== 'string') return '';
 
-	const decodedString = decodeURIComponent(stringToDecode);
+	let decodedString = '';
+
+	try {
+		decodedString = decodeURIComponent(stringToDecode);
+	} catch {
+		decodedString = stringToDecode;
```
mauriciokin, line 9: Would be the optimal expected behaviour to return the primary payload if it couldn't be decoded (catch error)?

packages/forms-web-app/src/lib/encode-string.js
```diff
@@ -1,7 +1,13 @@
 const encodeString = (stringToEncode) => {
 	if (!stringToEncode || typeof stringToEncode !== 'string') return '';
 
-	const encodedString = encodeURIComponent(stringToEncode);
+	let encodedString = '';
+
+	try {
+		encodedString = encodeURIComponent(stringToEncode);
+	} catch {
+		encodedString = stringToEncode;
```
mauriciokin, line 9: Would be the optimal expected behaviour to return the primary payload if it couldn't be decoded (catch error)? Doesn't it have a compounded effect when it goes through the rest of the steps?

packages/forms-web-app/src/lib/decode-string.js
christopher-semmens: This fixes an error when an invalid input was being decoded (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Malformed_URI). In user cases we still want the string but for the decode/encode to not take place. Looking into creating checks for valid strings looks quite cumbersome. This was the most elegant solution that I could find.

mauriciokin (approved):

mauriciokin (approved):
Legit

PR #363, opened 2022-07-28
fix: amend docker-compose.yml to run code from mounted volume

JamesP-KC:
- revert back to mounting local directories inside docker container
- switch to docker compose v2
- remove `common` from api docker container (not currently used)
- author change in package.json


mauriciokin (approved):
Legit!

christopher-semmens (approved):

PR #369, opened 2022-07-29
feat(hide examination timetable link): hide Examination Timetable Link

mauriciokin:
Hide Examination Timetable Link

This PR needs to be merged as the last of the hide link ones

### DIFF
--- a/packages/forms-web-app/src/views/layouts/vertical-tabs.njk
+++ b/packages/forms-web-app/src/views/layouts/vertical-tabs.njk
@@ -29,21 +29,27 @@
         </a>
       {% else %} Registration comments </br> (Relevant representations) {% endif %}
     </li>
-    <li class="side-menu govuk-body-s {% if 'timetable' == activeProjectLink %} active {% endif %}">
-      {% if 'timetable' != activeProjectLink %}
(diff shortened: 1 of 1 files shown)


packages/forms-web-app/src/controllers/projects/representations.js
```diff
@@ -62,7 +65,10 @@ exports.getRepresentation = async (req, res) => {
 			RepFrom: representation.data.RepFrom,
 			RepresentationRedacted: representation.data.RepresentationRedacted,
 			DateRrepReceived: representation.data.DateRrepReceived,
-			attachments: representation.data.attachments
+			attachments: representation.data.attachments,
+			hideAllExaminationDocumentsLink: true,
+			hideRecommendationAndDecisionLink: true,
+			hideExaminationTimetableLink: true
 		});
```
christopher-semmens, line 72: Hey @mauriciokin Would it maybe be an idea to add these 3 variables (if they always have the same values) to a config file which is imported by each of the controllers that use them?

mauriciokin: very good shout! fixed it!

christopher-semmens (approved):

PR #379, opened 2022-08-05
feat: if document type 'examination library' is present move to the s…

christopher-semmens:
…tart of the array


JamesP-KC (commented):
Do you reckon we can get a test added for this new logic? Perhaps something in `controllers/documents.test.js`?

mauriciokin (approved):
Legit

PR #383, opened 2022-08-10
fix(documents.njk): change in Published By to From on documents page

Philip-Philippou1:
Change of text in the documents page, 'Published By' becomes 'From'


e2e-tests/package-lock.json
```diff
@@ -1,7594 +1,28 @@
 {
 	"name": "e2e-tests",
 	"version": "1.0.0",
-	"lockfileVersion": 2,
+	"lockfileVersion": 1,
```
JamesP-KC, line 4: I think you can revert these changes to package-lock, there shouldn't be any changes needed if you are using node 16 / npm8

JamesP-KC: same with the other package-lock files, looks like they have been changed too. Try revert those changes, and i think the build should pass

PR #389, opened 2022-08-15
Filter Box - Relevant Representations Page requires correct case lettering

christopher-semmens:
### Task
[Filter Box - Relevant Representations Page requires correct case lettering](https://pins-ds.atlassian.net/jira/software/c/projects/ASB/boards/166?modal=detail&selectedIssue=ASB-507)


mauriciokin (approved):
Legit

PR #391, opened 2022-08-16
Feature/ASB-494 - Always show 'Application Documents' in document type filter

mauriciokin:
Always show 'Application Documents' in document type filter


packages/applications-service-api/src/utils/validate-parameters.js
```diff
@@ -0,0 +1,85 @@
+const validateParameters = (...args) => {
```
Philip-Philippou1, line 1: Is validateParameters and validateTypes used in the code?  I can only see them invoked in the test.

mauriciokin: I am adding this now to use it in the next tickets to replace the validation to all functions. Didn't want to use the functions now to make sure I won't get any bugs to add this kind of Type Checking Function

mauriciokin: All the functions are fully tested as well

Philip-Philippou1 (approved):

PR #394, opened 2022-08-17
fix(remove legend tag used inappropriately): remove legend tag used inappropriately

mauriciokin:
Remove legend tag used inappropriately

https://pins-ds.atlassian.net/jira/software/c/projects/ASB/boards/166?modal=detail&selectedIssue=ASB-349

### DIFF
--- a/packages/forms-web-app/src/views/cookies.njk
+++ b/packages/forms-web-app/src/views/cookies.njk
@@ -57,12 +57,9 @@
               <span class="govuk-hint">
                 <p>We use Google Analytics to measure how you use the application service so we can improve it based on user needs. We do not allow Google to use or share the data about how you use this site.</p>
                 <p>You agree to the Planning Inspectorate using your data to help improve the application service.</p>
+                <p><b>Do you want to accept analytics cookies?</b></p>
               </span>
               </div>
-
-              <legend class="govuk-fieldset__legend govuk-fieldset__legend--s">
-                Do you want to accept analytics cookies?
-              </legend>
 
               {{ govukRadios({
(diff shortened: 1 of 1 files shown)


packages/forms-web-app/src/views/cookies.njk
```diff
@@ -57,13 +57,10 @@
...
                 <p>We use Google Analytics to measure how you use the application service so we can improve it based on user needs. We do not allow Google to use or share the data about how you use this site.</p>
                 <p>You agree to the Planning Inspectorate using your data to help improve the application service.</p>
+                <p><b>Do you want to accept analytics cookies?</b></p>
               </span>
               </div>
 
-              <legend class="govuk-fieldset__legend govuk-fieldset__legend--s">
-                Do you want to accept analytics cookies?
-              </legend>
-
```
christopher-semmens, line 66: I believe this is the incorrect legend to be removed. This one needs to stay as is. If you check line 50 there is an h1 element wrapped in a legend. This is the one that needs to be removed. I would also switch out the h1 for a span just for good measure : )

mauriciokin: The thing is if that is done the order which the information is presented completely changes as the bottom legend becomes the top element and passes the previous h1 (new span) below it.

I also think this would change the ticket's AC itself.

christopher-semmens: The elements need to stay where they are but the first legend tag (nesting the h1) needs to be removed.

christopher-semmens (approved):

PR #403, opened 2022-08-24
Fix: ASB-636 - Pagination filtering issue

christopher-semmens:
### Task
https://pins-ds.atlassian.net/jira/software/c/projects/ASB/boards/166/backlog?view=detail&selectedIssue=ASB-636&issueLimit=100

### DIFF
--- a/packages/forms-web-app/__tests__/unit/utils/query-string-builder.test.js
+++ b/packages/forms-web-app/__tests__/unit/utils/query-string-builder.test.js
@@ -3,23 +3,25 @@ const { queryStringBuilder } = require('../../../src/utils/query-string-builder'
 describe('utils/query-string-builder', () => {
 	const valueToExpect = {
 		input1: {
-			params: null,
+			queryParams: null,
 			includedParams: []
 		},
 		input2: {
-			params: {},
-			includedParams: null
+			queryParams: {},
+			includedParams: null,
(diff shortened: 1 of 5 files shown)


packages/forms-web-app/__tests__/unit/utils/query-string-builder.test.js
```diff
@@ -3,64 +3,82 @@ const { queryStringBuilder } = require('../../../src/utils/query-string-builder'
...
 		input2: {
-			params: {},
-			includedParams: null
+			queryParams: {},
+			includedParams: null,
+			append: true
 		},
 		input3: {
-			params: {
+			queryParams: {
```
mauriciokin, line 15: The query param is what goes after the ? in the url. The param is what goes after the first slash. In this case this attribute should have a name of params in my opinion.

mauriciokin: Adding a console.log to application/forms in the controller documents.js file we can see the difference the queryParams normally have: page, type, category, stage attributes. The params have everything the query params have plus the info inside the slashes: caseRef, classification, page, type, category, stage.

mauriciokin (approved):

mauriciokin (approved):
lgtm

PR #407, opened 2022-08-30
feat(add what's your full name (myself)): Add What's your full name (myself) page

mauriciokin:
Add What's your full name (myself) page


packages/forms-web-app/src/controllers/examination/your-name.js
```diff
@@ -1,15 +1,46 @@
...
+
+	if (req.session && req.session.mySelfRegdata && body['full-name']) {
+		req.session.mySelfRegdata['full-name'] = body['full-name'];
+	}
+
+	if (req.query.mode === 'edit') {
+		res.redirect(`${ROUTE_PREFIX + YOUR_NAME}`);
+	} else {
+		res.redirect(`${ROUTE_PREFIX + YOUR_EMAIL_ADDRESS}`);
+	}
```
christopher-semmens, line 43: Hey @mauriciokin Just a suggestion that we set up the correct route and a placeholder page seeing as this is the first task. Then we can continue to work from that structure going forward. Let me know what you think though.

packages/forms-web-app/src/controllers/examination/your-name.js
```diff
@@ -1,15 +1,46 @@
...
+
+const setData = () => ({
+	backLinkUrl: WHO_ARE_YOU_SUBMITTING_FOR,
+	fullName: '',
+	title: 'What is your full name?'
+});
+
+const getYourName = async (req, res) => {
+	const { mySelfRegdata } = req.session;
+
```
christopher-semmens, line 15: @mauriciokin Do we need to store the data in different keys depending on the users route? For example, if we used one key, like 'examination' for example (`const { examination } = req.session;`) we could then define full name with the same key for every journey. We can use the value of 'who are you submitting for' for the logic.

mauriciokin: Done

christopher-semmens (approved):

PR #424, opened 2022-09-06
feat(add what's your interested party number): Add What's your interested party number controller, view and test

mauriciokin:
Add What's your interested party number controller, view and test


packages/forms-web-app/src/controllers/examination/your-interested-party-number.js
```diff
@@ -0,0 +1,71 @@
...
+				yourInterestedPartyNumber: {
+					view: yourInterestedPartyNumberView,
+					route: yourInterestedPartyNumberRoute
+				},
+				haveAnInterestedPartyNumber: { route: haveAnInterestedPartyNumberRoute },
+				submittingFor: { route: submittingForRoute }
+			}
+		}
+	},
+	routesConfig: { examination }
```
christopher-semmens, line 14: As this is only being used to extract the directory would it be better to add it to line 4 and declare it as 'examinationDirectory'? `directory: { examinationDirectory }` This would also be in keeping with the other controllers.

packages/forms-web-app/src/controllers/examination/your-interested-party-number.js
```diff
@@ -0,0 +1,71 @@
...
+			backLinkUrl,
+			errors,
+			errorSummary,
+			pageTitle,
+			title
+		});
+		return;
+	}
+
+	req.session[examinationSession.name].property = {
```
christopher-semmens, line 57: This would need to be `req.session[examinationSession.name][examinationSession.property. interestedPartyNumber]`

packages/forms-web-app/src/controllers/examination/your-interested-party-number.js
```diff
@@ -0,0 +1,71 @@
...
+				submittingFor: { route: submittingForRoute }
+			}
+		}
+	},
+	routesConfig: { examination }
+} = require('../../routes/config');
+
+const config = require('../../config');
+
+const interestedPartyNumberString = 'interested-party-number';
```
christopher-semmens, line 19: It would be good if this was set to the id retrieved from the '/routes/config' 'yourInterestedPartyNumber' object. We would then be able to feed that into the view and use it there.

packages/forms-web-app/src/controllers/examination/your-interested-party-number.js
mauriciokin: it was not being used so I had to remove it for lint purposes

packages/forms-web-app/src/controllers/examination/your-interested-party-number.js
```diff
@@ -0,0 +1,72 @@
...
+		return;
+	}
+
+	req.session[examinationSession.name][examinationSession.property.interestedPartyNumber] =
+		body[yourInterestedPartyNumberId];
+
+	if (req.query.mode === 'edit') {
+		res.redirect(`${examination.directory + yourInterestedPartyNumberRoute}`);
+	} else {
+		console.error('shouldEndHere');
```
christopher-semmens, line 64: Remove `console.error`

packages/forms-web-app/src/validators/shared/index.js
```diff
@@ -0,0 +1,17 @@
+const { body } = require('express-validator');
```
christopher-semmens, line 1: File name is `index.js`, do we want this? Maybe something like `not-empty-and-length.js`?

packages/forms-web-app/src/views/pages/examination/your-interested-party-number.njk
```diff
@@ -0,0 +1,52 @@
...
+					errorMessage: errors[id] and {
+						text: errors[id].msg
+					}
+				}) }}
+			</div>
+
+			{{ govukButton({
+				text: "Continue",
+				type: "Submit",
+				attributes: { "data-cy":"button-save-and-continue"}
```
christopher-semmens, line 47: 'button-save-and-continue' isn't an accurate description. Maybe something like 'button-submit-and-continue'?

packages/forms-web-app/src/controllers/examination/your-interested-party-number.js
```diff
@@ -0,0 +1,72 @@
...
+		id: yourInterestedPartyNumberId,
+		backLinkUrl,
+		interestedPartyNumber,
+		pageTitle,
+		title
+	});
+};
+
+const postYourInterestedPartyNumber = (req, res) => {
+	const { body = { yourInterestedPartyNumberId: '' } } = req;
```
christopher-semmens, line 43: Could have a try/catch here to handle any errors.

packages/forms-web-app/src/validators/shared/index.js
mauriciokin: from a compiler perspective it wouldn't make much difference... I will create both files and export them from the index then.

packages/forms-web-app/src/controllers/examination/your-interested-party-number.js
mauriciokin: added for both get and post routes

christopher-semmens (approved):

PR #438, opened 2022-09-16
ASB-575: Are you the applicant?

christopher-semmens:
### Task
https://pins-ds.atlassian.net/jira/software/c/projects/ASB/boards/166?modal=detail&selectedIssue=ASB-575

### DIFF
--- a/packages/forms-web-app/__tests__/unit/controllers/examination/has-interested-party-number.test.js
+++ b/packages/forms-web-app/__tests__/unit/controllers/examination/has-interested-party-number.test.js
@@ -4,33 +4,14 @@ const {
 } = require('../../../../src/controllers/examination/has-interested-party-number');
 const { mockReq, mockRes } = require('../../mocks');
 
-const {
-	routesConfig: {
-		examination: {
-			directory: examinationDirectory,
-			pages: {
-				applicant: { route: applicantRoute },
-				hasInterestedPartyNumber: {
-					options: {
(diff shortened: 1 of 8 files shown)


packages/forms-web-app/__tests__/unit/controllers/examination/has-interested-party-number.test.js
```diff
@@ -156,16 +142,16 @@ describe('controllers/examination/has-interested-party-number', () => {
...
 				...req,
 				body: {
-					['examination-has-interested-party-number']: 'no'
+					'examination-has-interested-party-number': 'no'
 				}
 			};
 
 			postHasInterestedPartyNumber(mockRequest, res);
-			expect(res.redirect).toHaveBeenCalledWith(`${examinationDirectory + applicantRoute}`);
+			expect(res.redirect).toHaveBeenCalledWith('/examination/are-you-applicant');
```
mauriciokin, line 154: Not better to keep the config route so when they get changed the test still works? I know this can be a downside in case it gets set to an improper path but it should be noticed on the review or at latest on dev.

christopher-semmens: @mauriciokin Reasoning for hardcoding values in test is to avoid unintentional changes in the config as any changes will be flagged in failing tests.

mauriciokin (approved):

PR #441, opened 2022-09-20
ASB-689: Update OpenAPI spec with Submissions API

JamesP-KC:
- Document new POST submissions API endpoint in the OpenAPI spec

### DIFF
--- a/packages/applications-service-api/api/openapi.yaml
+++ b/packages/applications-service-api/api/openapi.yaml
@@ -1044,12 +1044,11 @@ components:
       properties:
         file:
           type: string
-          description: File to be uploaded
+          description: File object to be uploaded. **Required unless submission is a comment**
           format: binary
       allOf:
         - $ref: '#/components/schemas/SubmissionProperties'
       required:
-        - file
         - name
         - email
         - interestedParty
@@ -1059,13 +1058,6 @@ components:
       allOf:
         - $ref: '#/components/schemas/SubmissionProperties'
       properties:
-        caseReference:
-          type: string
-          example: EN010118
-        dateSubmitted:
-          type: string
-          format: date
-          example: 2022-08-02 08:24:05
         file:
           type: object
           properties:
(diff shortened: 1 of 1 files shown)


mauriciokin (approved):
legit

packages/applications-service-api/api/openapi.yaml
```diff
@@ -1006,4 +1039,97 @@ components:
           example: 1
         currentPage:
           type: number
-          example: 1
\ No newline at end of file
+          example: 1
+    SubmissionCreateRequestBody:
+      properties:
+        file:
```
Philip-Philippou1, line 1045: Do we want an MD5/size to be sent in the request as a way of validating the file at both ends of transfer.  Perhaps we don't.

packages/applications-service-api/api/openapi.yaml
```diff
@@ -1006,4 +1039,97 @@ components:
...
+          description: Interested party reference number
+          example: 3287498209
+        deadline:
+          type: string
+          description: Name/title of Deadline
+          example: Deadline 1A
+        submissionType:
+          type: string
+          example: "Other: Biodiversity Ecology and natural Environment"
+        representation:
```
Philip-Philippou1, line 1110: Is the idea that if it's a comment we would send it using this endpoint as text and then the API can convert to PDF? 
That would work but we may need to remove the 'required' indicator from file.

JamesP-KC: Yes good call, I have pushed an update making `representation` and `file` optional but have added a description to try and highlight that one of them is required. There doesn't seem to be a non-hacky way of defining such requirements in OpenAPI

packages/applications-service-api/api/openapi.yaml
```diff
@@ -1006,4 +1039,97 @@ components:
...
+          type: string
+          description: Interested party reference number
+          example: 3287498209
+        deadline:
+          type: string
+          description: Name/title of Deadline
+          example: Deadline 1A
+        submissionType:
+          type: string
+          example: "Other: Biodiversity Ecology and natural Environment"
```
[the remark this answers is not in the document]
JamesP-KC: Strange, it was rendering OK for me in editor.swagger.io. In any case, I've changed it to another example without quotes :)

Philip-Philippou1 (approved):

PR #443, opened 2022-09-22
fix(declaration.js): ASB-620 handle internal exception when posting declaration

Philip-Philippou1:
Add exception handling to relevant reps - POST to avoid uncaught exception forcing a web server re-start.

This is a generic catch all for now which will result in the same outcome on a 500 error from API - Page-not found.  The change here handles the exception which will avoid an uncaught exception and resulting web-app restart.

We may need to revisit this in future as error handling in this area requires some work.


JamesP-KC:
Looks good to me 👍

JamesP-KC (approved):

PR #444, opened 2022-09-22
Feature/ASB-580 - What's your email address?

mauriciokin:


packages/forms-web-app/src/controllers/examination/email.js
```diff
@@ -2,24 +2,61 @@ const {
...
 	id: email.id,
 	pageTitle: email.name,
-	title: email.name
+	title: email.name,
+	hint: "We'll use your email address to confirm we've received your submission. We will not publish your email address."
 };
 
 const getEmail = async (req, res) => {
+	const requestSession = req.session;
+	const userEmail = requestSession?.[examinationSessionStorage.name]?.userEmail;
```
christopher-semmens, line 22: looks like 'userEmail' doesn't need to be used until after the conditional '(!requestSession || !currentView || !currentView?.route)'. Would it be better to declare and set it after?

packages/forms-web-app/src/controllers/examination/email.js
```diff
@@ -2,24 +2,61 @@ const {
...
+	const currentView = requestSession?.currentView;
+	userEmail && (pageData.email = userEmail);
+
+	if (!requestSession || !currentView || !currentView?.route)
```
christopher-semmens, line 26: We would always have a req.session so probably no need to check for it.

packages/forms-web-app/src/controllers/examination/email.js
```diff
@@ -2,24 +2,61 @@ const {
...
 	pageTitle: email.name,
-	title: email.name
+	title: email.name,
+	hint: "We'll use your email address to confirm we've received your submission. We will not publish your email address."
 };
 
 const getEmail = async (req, res) => {
+	const requestSession = req.session;
+	const userEmail = requestSession?.[examinationSessionStorage.name]?.userEmail;
+	const currentView = requestSession?.currentView;
```
christopher-semmens, line 23: What is current view doing?

packages/forms-web-app/src/controllers/examination/email.js
```diff
@@ -2,24 +2,61 @@ const {
...
+	userEmail && (pageData.email = userEmail);
+
+	if (!requestSession || !currentView || !currentView?.route)
+		return res.status(404).render('error/not-found');
+
+	pageData.backLinkUrl = `${examinationDirectory + currentView.route}`;
```
christopher-semmens, line 29: Can be written with out the quotations: `pageData.backLinkUrl = examinationDirectory + currentView.route;`

packages/forms-web-app/src/controllers/examination/email.js
```diff
@@ -2,24 +2,61 @@ const {
...
 	res.render(email.view, pageData);
 };
 
 const postEmail = async (req, res) => {
-	res.render(email.view, pageData);
+	const { body = {} } = req;
+	const { errors = {}, errorSummary = [] } = body;
+	const requestSession = req?.session;
+	const currentView = requestSession?.currentView;
+	const examinationStorage = requestSession?.[examinationSessionStorage.name];
```
christopher-semmens, line 38: May be best to check for the examination session first, without it we should not carry on:

```
const examinationStorage = req.session[examinationSessionStorage.name];

if (!examinationStorage) ....
```

packages/forms-web-app/src/controllers/examination/email.js
```diff
@@ -2,24 +2,61 @@ const {
...
+
+	if (errors[email.id] || Object.keys(errors).length > 0) {
+		return res.render(email.view, {
+			...pageData,
+			errors,
+			errorSummary
+		});
+	}
+
+	if (!requestSession || !currentView) return res.status(404).render('error/not-found');
```
christopher-semmens, line 50: As mentioned before, we would always have the 'requestSession' but what to check for the req.session.examination

packages/forms-web-app/src/controllers/examination/email.js
```diff
@@ -2,24 +2,61 @@ const {
...
+
+	examinationStorage.userEmail = examinationSessionStorage.property.email = body[email.id];
+	pageData.backLinkUrl = `${examinationDirectory + currentView.route}`;
```
christopher-semmens, line 53: This would want to be set before the errors conditional as we would still want the backlink on that render

packages/forms-web-app/src/controllers/examination/email.js
```diff
@@ -2,24 +2,61 @@ const {
...
+
+	if (req.query?.mode === 'edit') {
+		res.render(email.view, pageData);
```
christopher-semmens, line 56: I know this is 'placeholder' at the moment but we have a route for 'Check your answers' that will be used here

packages/forms-web-app/src/controllers/examination/email.js
mauriciokin: it breaks the tests if you forget to pass it and in case someone gets to the link in an anonymous tab they should get a 404

packages/forms-web-app/src/controllers/examination/email.js
mauriciokin: cool will fix it

packages/forms-web-app/src/controllers/examination/email.js
mauriciokin: it passes the name object (agent, organization, onBehalf) from the route config through the forward-view middleware in the name.js (as we have 3 different names using the same controller).

mauriciokin: knowing which is the right view we can create the proper backlink to it

packages/forms-web-app/src/controllers/examination/email.js
mauriciokin: It can but there is a caveat... With string interpolation we always guarantee JS will cast the variable values to a string. This doesn't happen with a sum (+) operator.

Eg:

`
const x = undefined;
const y = ' ';

console.log(typeof x + y) = undefined 
console.log(typeof `${x + y}`) = string
`

packages/forms-web-app/src/controllers/examination/email.js
mauriciokin: If someone sends a link to a friend for example and they open in the middle of the journey they should get a not found message as they don't have a session.

packages/forms-web-app/src/controllers/examination/email.js
mauriciokin: makes sense

packages/forms-web-app/src/controllers/examination/email.js
mauriciokin: Or when they pass 'edit' they wanna go straight to the 'Check your answers' all the time?

packages/forms-web-app/src/controllers/examination/email.js
mauriciokin: It can but there is a caveat... With string interpolation we always guarantee JS will cast the variable values to a string. This doesn't happen with a sum (+) operator.

Eg:

`
const x = undefined;
const y = ' ';

console.log(typeof x + y) = undefined
console.log(typeof ${x + y}) = string
`

packages/forms-web-app/src/controllers/examination/email.js
christopher-semmens: They will always have a session (`req.session`) but we want to check if they have started the examination journey (`req.session.examination`)

packages/forms-web-app/src/controllers/examination/email.js
christopher-semmens: Sorry, I'm not describing the issue so well. The first check we would want to do is to check for an 'examination' session as without this we can assume that the journey has not been started and we also don't have a key for our data. This is why it should be declared first and for the sake of optimisation, before any other logic.

mauriciokin: fixed it

packages/forms-web-app/src/controllers/examination/email.js
mauriciokin: this was raised by Vimal while testing

packages/forms-web-app/src/controllers/examination/email.js
christopher-semmens: As mentioned below, the user would always have an req.session as it is applied to Express.js. We want to check for req.session.examination

packages/forms-web-app/src/controllers/examination/name.js
```diff
@@ -34,10 +43,13 @@ const postName = async (req, res) => {
...
 	const examinationSession = session?.[examinationSessionStorage.name];
+	const sessionCurrentView = req.session?.currentView;
 
-	if (!examinationSession || !req.currentView?.id) return res.status(404).render('error/not-found');
+	if (!examinationSession || !sessionCurrentView?.id)
+		return res.status(404).render('error/not-found');
 
-	const { id, view } = req.currentView;
+	const { id, view } = sessionCurrentView;
+	examinationSessionStorage.property.name = req.body[id] ?? '';
```
christopher-semmens, line 52: Is this correct? I think this should be `examinationSession[examinationSessionStorage.property.name] = req.body[id] ?? '';`

packages/forms-web-app/src/controllers/examination/email.js
```diff
@@ -2,24 +2,67 @@ const {
...
 
 const postEmail = async (req, res) => {
-	res.render(email.view, pageData);
+	const { body = {} } = req;
+	const { errors = {}, errorSummary = [] } = body;
+	const requestSession = req?.session;
+	const currentView = requestSession?.currentView;
+	const examinationStorage = requestSession?.[examinationSessionStorage.name];
+
+	if (!requestSession || !currentView || !examinationStorage)
```
christopher-semmens, line 45: We do not need to check for a `requestSession` here.

packages/forms-web-app/src/controllers/examination/email.js
```diff
@@ -2,24 +2,67 @@ const {
...
+		return res.status(404).render('error/not-found');
+
+	examinationStorage?.userEmail && (pageData.email = examinationStorage?.userEmail);
```
christopher-semmens, line 48: `.userEmail` should be extracted from the config variables.

packages/forms-web-app/src/controllers/examination/email.js
```diff
@@ -2,24 +2,67 @@ const {
...
+
+	examinationStorage.userEmail = examinationSessionStorage.property.email = body[email.id];
```
christopher-semmens, line 50: I think this should be `examinationStorage.userEmail = examinationStorage[examinationSessionStorage.property.email] = body[email.id];`

`.userEmail` should also be extracted from the config variables

mauriciokin: Probably what we want is examinationStorage.email

this: examinationSessionStorage.property.email will be transformed in the email value but once it gets reassigned it can override other variables

examinationStorage  = req.session.examination

examinationSessionStorage = config.sessionStorage.examination

examinationStorage[examinationSessionStorage.property.email] = sessionStorage.examination.email (this is wrong)

christopher-semmens (approved):

PR #451, opened 2022-09-26
Task/ASB-747 Address line 3 becomes Town or county

Philip-Philippou1:
Address Line 3 label becomes Town or county in the address entry.  
This commit also fixes a large number of tests due to the click location problem for over 100 tests.

### DIFF
--- a/e2e-tests/cypress/support/common-methods/clickProjectLink.js
+++ b/e2e-tests/cypress/support/common-methods/clickProjectLink.js
@@ -1,6 +1,6 @@
 module.exports = (projectName) => {
 	cy.get('td:nth-child(1)').each(($e1, index) => {
-		const text = $e1.text().trim();
+		const text = $e1.text();
 		if (text.includes(projectName)) {
 			cy.get('td:nth-child(1)').eq(index).click(15, 20);
 		}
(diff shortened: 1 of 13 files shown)


e2e-tests/cypress/support/common-methods/clickProjectLink.js
```diff
@@ -2,7 +2,7 @@ module.exports = (projectName) => {
 	cy.get('td:nth-child(1)').each(($e1, index) => {
 		const text = $e1.text();
 		if (text.includes(projectName)) {
-			cy.get('td:nth-child(1)').eq(index).click();
+			cy.get('td:nth-child(1)').eq(index).click(15, 20);
```
JamesP-KC, line 5: I think the issue here is that the query selector is getting the `td` not the actual `a` within it. Clearly up until now, that has worked but i wonder if revising the selector is the "proper" fix.

I did a quick test with this and seemed to work:
```
module.exports = (projectName) => {
	cy.get('td:nth-child(1)').contains(projectName).click();
};
```

JamesP-KC (approved):

e2e-tests/cypress/support/common-methods/clickProjectLink.js
Philip-Philippou1: This works great, will use, thanks

JamesP-KC (approved):

PR #460, opened 2022-10-04
feat(make a comment): add Make a comment page, add router, controller, view for make a comment

jonathangoulding:


.vscode/launch.json
```diff
@@ -0,0 +1,19 @@
+{
+	// Use IntelliSense to learn about possible attributes.
```
jonathangoulding, line 2: Can add this use to the readme ? 

Or is it committed by mistake ?

.vscode/launch.json
```diff
@@ -0,0 +1,19 @@
...
+	"version": "0.2.0",
+	"configurations": [
+		{
+			"name": "Docker: Attach to Node",
+			"type": "node",
+			"request": "attach",
+			"port": 9229,
+			"address": "localhost",
+			"localRoot": "${workspaceFolder}",
+			"restart": true,
```
jonathangoulding, line 14: this allow nodemon to be run under the dev debug command in the docker-compose

packages/forms-web-app/__tests__/unit/controllers/examination/enter-comment.test.js
```diff
@@ -2,43 +2,179 @@ const {
 	getEnterComment,
 	postEnterComment
 } = require('../../../../src/controllers/examination/enter-comment');
-const { mockReq, mockRes } = require('../../mocks');
+const { mockReq, mockRes, mockResponse } = require('../../mocks');
+
+const pathShortner = (obj, path) => path.split('.').reduce((value, el) => value && value[el], obj);
```
jonathangoulding, line 7: would this be better as: previousValue, currentValue

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce

packages/forms-web-app/__tests__/unit/controllers/examination/enter-comment.test.js
```diff
@@ -2,43 +2,179 @@ const {
...
+
+			getEnterComment(mockRequest, res);
+
+			expect(res.render).toHaveBeenCalledWith('pages/examination/enter-comment', {
+				...pageData,
+				hint: eval(hintValue)
+			});
+		});
+
+		it('should render the view with default and session pageData', function () {
```
jonathangoulding, line 62: should be arrow function

packages/forms-web-app/__tests__/unit/controllers/examination/enter-comment.test.js
```diff
@@ -2,43 +2,179 @@ const {
...
+	id: 'examination-enter-a-comment',
+	pageTitle: 'Make a comment',
+	title: 'Make a comment',
+	hint: 'Any further information requested by',
+	captionTitle: 'Deadline item:',
+	comment: ''
 };
 
-describe('controllers/examination/submitting-for', () => {
+const hintValue = '`${pageData.hint} ${pathShortner(mockRequest, itemsPath)[0].submissionItem}`';
```
jonathangoulding, line 20: this the same after eval for each hint. 

Can just be set to: "Any further information requested by Statement of Commonality of SoCG "

and remove this complexity

packages/forms-web-app/__tests__/unit/controllers/examination/enter-comment.test.js
```diff
@@ -2,43 +2,179 @@ const {
...
 
-	describe('postSubmittingFor', () => {
-		it('should call the correct template', () => {
+	describe('postEnterComment', () => {
+		it('should redirect user to /examination/comment-has-personal-information-or-not', () => {
+			const mockRequest = { ...req };
+			postEnterComment(mockRequest, res);
+
+			expect(res.render).not.toHaveBeenCalled();
+			expect(pathShortner(mockRequest, itemsPath)[0].submissionType);
```
jonathangoulding, line 96: These don't appear to test anything or assert in a way that suggest the test is worth doing. 

Change to `expect(pathShortner(mockRequest, itemsPath));` and the test still passes ?

packages/forms-web-app/__tests__/unit/controllers/examination/enter-comment.test.js
jonathangoulding: Or this would no be necessary as comments below

packages/forms-web-app/__tests__/unit/controllers/examination/enter-comment.test.js
```diff
@@ -2,43 +2,179 @@ const {
...
-			expect(res.render).toHaveBeenCalledWith('pages/examination/enter-comment', pageData);
+			expect(res.render).toHaveBeenCalledWith('error/not-found');
+			expect(res.redirect).not.toHaveBeenCalled();
 		});
 	});
 
-	describe('postSubmittingFor', () => {
-		it('should call the correct template', () => {
+	describe('postEnterComment', () => {
+		it('should redirect user to /examination/comment-has-personal-information-or-not', () => {
```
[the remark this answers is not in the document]
mauriciokin: Chris wants the test paths hardcoded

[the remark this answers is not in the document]
mauriciokin: I would rather have them all from the config in all tests and create a new test to check if the config routes are as expected in its own test...

packages/forms-web-app/__tests__/unit/controllers/examination/enter-comment.test.js
mauriciokin: done

packages/forms-web-app/__tests__/unit/controllers/examination/enter-comment.test.js
mauriciokin: forgot to add the assertion at the end

packages/forms-web-app/__tests__/unit/controllers/examination/enter-comment.test.js
mauriciokin: fixed

packages/forms-web-app/__tests__/unit/controllers/examination/enter-comment.test.js
mauriciokin: I can modify it to match this convention but as reduce parameters are done in order a dev should know what they mean

.vscode/launch.json
mauriciokin: Done a little googling and it seems to be useful info I would keep it.

packages/forms-web-app/src/controllers/examination/enter-comment.js
```diff
@@ -1,28 +1,89 @@
...
+	examinationSession[sessionIdPrimary]?.[sessionIdTertiary]?.[
+		examinationSession[sessionIdPrimary]?.[sessionIdSecondary]
+	];
+
+const getEnterComment = async (req, res) => {
+	const { session = {} } = req;
+	const examinationSession = session?.[examinationSessionStorage.name];
+
+	if (!examinationSession || !examinationSession[sessionIdPrimary])
+		return res.status(404).render('error/not-found');
```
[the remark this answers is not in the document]
mauriciokin: Not sure it is a great idea be throwing errors there and if there is an error boundary middleware that would gracefully handle them afterwards in their specific case.

We can verify it and if so we can raise a ticket to change it but not do this change in this one as it would break the pattern through all controllers we made previously and this change is out of the scope of this ticket itself...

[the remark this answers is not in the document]
mauriciokin: Would also be good to if we could find a solution to the jest mock for chainable functions this would be awesome actually.

packages/forms-web-app/__tests__/unit/controllers/examination/enter-comment.test.js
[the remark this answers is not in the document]
jonathangoulding: wrong line i meant for the redirect to to have been called with but i agree they should be hard coded.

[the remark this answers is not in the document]
mauriciokin: I would have them hard coded in a config test and dynamic in all other tests its much better.

christopher-semmens (approved):

PR #467, opened 2022-10-11
feat(api): integrate submission endpoint with ni

JamesP-KC:
- adds integration with NI api `submissionsupload/file` endpoint, to upload file
- inserts submission data (including file metadata, if file is uploaded) to submissions table


packages/applications-service-api/src/controllers/submissions.js
```diff
@@ -1,47 +1,76 @@
...
-		} else {
+		if (!req.body.representation && !req.file) {
 			return res.status(StatusCodes.BAD_REQUEST).send({
 				code: StatusCodes.BAD_REQUEST,
 				errors: ["Missing property 'representation' or 'file'"]
 			});
 		}
 
-		return res.status(StatusCodes.CREATED).send(responseBody);
+		const { caseReference } = req.params;
```
Philip-Philippou1, line 16: It looks like a comment and file would be supported with one entry in the DB for both.  That may be fine but something to think about in the future ticket where we will need to create a PDF of comment and I think have a separate DB entry.  Think this is ok for now though.

packages/applications-service-api/src/controllers/submissions.js
```diff
@@ -1,47 +1,76 @@
...
+	});
+
+	return {
+		...submission,
+		file: fileData
+	};
+};
+
+const buildSubmissionData = (requestBody, caseReference) => {
+	return {
```
Philip-Philippou1, line 58: Some of the fields are marked as mandatory in the API.  Was wondering if we should try and enforce it.

JamesP-KC: I was thinking that was more an 'unhappy path' type thing to do. But actually we don't have that covered in the forthcoming ticket so I will implement it now

packages/applications-service-api/src/controllers/submissions.js
JamesP-KC: Looks like we already have ASB-801. I've added a couple of scenarios to try and distinguish the behaviour when submitting both a file and a comment

packages/applications-service-api/__tests__/unit/controllers/submissions.test.js
```diff
@@ -1,86 +1,131 @@
...
+		const fileData = {
+			name: 'Test-123-1.png',
+			originalName: 'Test.png',
+			size: 83,
+			md5: '3ac9e57c0901b59075291537496aaf06'
+		};
+
+		afterEach(() => jest.resetAllMocks());
+
+		it('should return file name including submissionId and sequence number if file uploaded', async () => {
```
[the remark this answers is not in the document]
JamesP-KC: Could be clearer, but that is contained within fileData:
https://github.com/Planning-Inspectorate/applications-service/pull/467/files/4b42c0cef1efc8853639ece6d635ee46d2de6e5f#diff-d4f1d8a9142b34b54ec3a1a506d853d3550c16c9efe4f8f31194b2977a2060f8R90
https://github.com/Planning-Inspectorate/applications-service/pull/467/files/4b42c0cef1efc8853639ece6d635ee46d2de6e5f#diff-d4f1d8a9142b34b54ec3a1a506d853d3550c16c9efe4f8f31194b2977a2060f8R63

packages/applications-service-api/src/services/submission.service.js
```diff
@@ -0,0 +1,60 @@
...
+		submissionType: submission.submissionType,
+		sensitiveData: submission.sensitiveData,
+		lateSubmission: submission.lateSubmission,
+		caseReference: submission.caseReference,
+		dateSubmitted: submission.dateSubmitted,
+		submissionId: submissionId,
+		representation: submission.representation
+	});
+
+	if (submissionId === PLACEHOLDER_SUBMISSION_ID) {
```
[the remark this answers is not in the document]
JamesP-KC: The behaviour I saw when testing from an empty DB locally was that if the submissions table was empty, the first row should be given an ID of `1` rather than `0`.

Philip-Philippou1 (approved):

PR #476, opened 2022-10-21
feat(personal information which): add personal information which page

jonathangoulding:
add switch for file and both journey

add tests

introduce concept of component tests

introduce folder over file structure


packages/forms-web-app/src/controllers/examination/personal-information-which/utils/savePersonalInformationFlags.js
```diff
@@ -0,0 +1,35 @@
+const { makeIntoArray } = require('../../select-file/utils/helpers');
```
jonathangoulding, line 1: Shall we move this file into session under - personal info ?

christopher-semmens: I think it would make sense where it currently is as it is specific to this page.

christopher-semmens (approved):

PR #485, opened 2022-11-03
Add another deadline item

jonathangoulding:


packages/forms-web-app/src/config.js
```diff
@@ -66,8 +66,10 @@ module.exports = {
 				description: 'description',
 				email: 'email',
 				hasInterestedPartyNo: 'hasInterestedPartyNo',
+				comment: 'comment',
 				id: 'id',
```
christopher-semmens, line 70: Are these values being used. Are we not moving away from this config?

christopher-semmens (approved):

packages/forms-web-app/src/config.js
jonathangoulding: addressed in next PR for removal

PR #486, opened 2022-11-04
Task/asb 876 sanitise role and name

Philip-Philippou1:
Add sanitisation to full-name and job-role forms.


packages/forms-web-app/src/controllers/register/myself/full-name.js
```diff
@@ -1,25 +1,46 @@
 const { VIEW } = require('../../../lib/views');
 
+function sanitiseFormPostResponse(error, url) {
```
Philip-Philippou1, line 3: Might want to pull this out of here as it's replicated in 4 places, just wasn't sure where to put it and it's only 3 lines.

JamesP-KC: perhaps `src/utils`?

also, using arrow function syntax would be preferable imo
```
const sanitiseFormPostResponse = (error, url) => {
    // implementation
};
```

JamesP-KC: i'm not actually clear what that function is doing. The name implies it is sanitising something, but just assigns the params to `this` (not sure what `this` refers to in this context).

packages/forms-web-app/src/views/register/organisation/full-name.njk
```diff
@@ -4,6 +4,13 @@
 {% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
 {% from "govuk/components/input/macro.njk" import govukInput %}
 
+{% block script %}
+	{% set initiateScriptsConfig = (initiateScriptsConfig.push({
```
JamesP-KC, line 8: what will happen if the user doesn't have JS enabled?

packages/forms-web-app/src/controllers/register/myself/full-name.js
JamesP-KC: OK i looked a bit more carefully and think i understand, it's a constructor. So arrow function not appropriate!

packages/forms-web-app/src/views/register/organisation/full-name.njk
JamesP-KC: also perhaps rather than repeating x3, this block could be defined in a separate njk file and imported into each of the 3 page templates

JamesP-KC:
This feels like something that should be solved at the firewall level, as it is a false-positive. I can't think of a solution to this problem that doesn't involve using JavaScript on the frontend.

packages/forms-web-app/src/controllers/register/myself/full-name.js
Philip-Philippou1: Good shout, I have moved this to utils and replaced the 8 or so instances with an import.

packages/forms-web-app/src/views/register/organisation/full-name.njk
Philip-Philippou1: Seems ok without javascript, just doesn't run the script from what I can see in browser debug.  Seems to work ok, we have a general problem with back button but that is not to do with this change.

I've been trying to move the script out of the njk into it's own njk file for import.  It basically fails because the initiateScript is not available -
Error: Unable to call `initiateScriptsConfig["push"]`, which is undefined or falsey

Think it's to do with the order of imports of the templates and which script are available at run time.  I may leave this one.

JamesP-KC (approved):

PR #488, opened 2022-11-04
ASB-801: upload submission comment as pdf file to ni

JamesP-KC:
- adds functionality to generate a PDF file containing the comment and upload it to NI
- add validation of max length/size of `representation` (<= 65234 chars)

### DIFF
--- a/packages/applications-service-api/src/controllers/submissions.js
+++ b/packages/applications-service-api/src/controllers/submissions.js
@@ -1,6 +1,6 @@
 const { StatusCodes } = require('http-status-codes');
 const { createSubmission } = require('../services/submission.service');
-const { submitFile } = require('../services/ni.file.service');
+const { submitUserUploadedFile, submitRepresentationFile } = require('../services/ni.file.service');
 
 module.exports = {
 	async createSubmission(req, res) {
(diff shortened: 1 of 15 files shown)


packages/applications-service-api/src/controllers/submissions.js
```diff
@@ -10,7 +10,11 @@ module.exports = {
 		let submission = await createSubmission(submissionRequestData);
 
 		if (req.file) {
-			submission = await submitFile(submission, req.file);
+			submission = await submitUserUploadedFile(submission, req.file);
+		}
+
+		if (submission.representation) {
+			submission = await submitRepresentationFile(submission);
```
Philip-Philippou1, line 17: We are exposing details of the file that we create in the API response, I think that's fine, but it may go away in the future so probably want to be careful the FE doesn't build functionality against this.   I think it's fine though.

Philip-Philippou1 (dismissed):

packages/applications-service-api/src/controllers/submissions.js
JamesP-KC: Good point, I'll remove `file` from the response (if it's a submission with `representation`).

Philip-Philippou1 (approved):

PR #493, opened 2022-11-10
fix(representations.js): truthy check for representations

RaySage:


JamesP-KC (commented):
I think the error is occurring in the `getRepresentation` function, rather than `getRepresentations`. So not sure this change will have an effect. I've documented my findings on the ticket - https://pins-ds.atlassian.net/browse/ASB-855?focusedCommentId=55444

RaySage:
Thanks @JamesP-KC , sorry I'm not too familiar with the functionality of the APIs, so something like below,
if the return of the unhandled-exception page is not there then the request times out or never completes, so not very user friendly. Could have the check for not 200 immediately after the getRepresentation() of course, depending on preference.
And the same for getRepresentations() as well

`exports.getRepresentation = async (req, res) => {
	const applicationResponse = await getAppData(req.params.case_ref);
	if (applicationResponse.resp_code === 200) {
		const representation = await getRepresentation(req.params.id);
		if (representation.resp_code === 200) {
			res.render(VIEW.PROJECTS.REPRESENTATION, {
				projectName: applicationResponse.data.ProjectName,
				caseRef: applicationResponse.data.CaseReference,
				hideProjectInformationLink,
				hideAllExaminationDocumentsLink,
				hideRecommendationAndDecisionLink,
				hideExaminationTimetableLink,
				RepFrom: titleCase(representation.data.RepFrom),
				PersonalName: representation.data.PersonalName,
				RepresentationRedacted: representation.data.RepresentationRedacted,
				DateRrepReceived: representation.data.DateRrepReceived,
				attachments: representation.data.attachments
			});
		} else {
			return res.status(500).render('error/unhandled-exception');
		}
	}
};`

Philip-Philippou1 (commented):
I think James mentioned this is a not found issue so we probably want to check a 404 first

if (representation.resp_code === 404) {
  return res.status(404).render('error/not-found');

JamesP-KC (approved):

PR #503, opened 2022-11-17
feat(applications-service-api): complete submission endpoint

JamesP-KC:
- new `/submissions/{submissionId}/complete` endpoint
  - updated openapi spec
  - stub implementation
      - request validation
      - fetching submission from DB
      - does not call notify service yet, to be implemented in ASB-901
 - fix incorrect path of create submission endpoint. `/submission/:caseReference` -> `/submissions/:caseReference`

### DIFF
--- a/packages/applications-service-api/src/error/apiError.js
+++ b/packages/applications-service-api/src/error/apiError.js
@@ -10,8 +10,12 @@ class ApiError {
 		return new ApiError(StatusCodes.BAD_REQUEST, { errors: messages.flat() });
 	}
 
+	static notFound(message) {
+		return new ApiError(StatusCodes.NOT_FOUND, { errors: [message] });
+	}
+
 	static applicationNotFound(id) {
-		return new ApiError(StatusCodes.NOT_FOUND, { errors: [`Application ${id} was not found`] });
+		return this.notFound(`Application ${id} was not found`);
 	}
 
 	static applicationNotAcceptable(id) {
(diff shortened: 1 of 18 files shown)


packages/applications-service-api/src/error/apiError.js
```diff
@@ -21,39 +25,27 @@ class ApiError {
...
-			errors: [`Interested party for project ${caseRef} not found`]
-		});
+		return this.notFound(`Interested party for project ${caseRef} not found`);
 	}
 
 	static noDocumentsFound() {
-		return new ApiError(404, { errors: [`No documents found`] });
-	}
-
-	static commentsForPartyWithIDNotUpdated(ID) {
```
Philip-Philippou1, line 37: Is this a not found or could something else result in this error?

packages/applications-service-api/src/error/apiError.js
```diff
@@ -10,8 +10,12 @@ class ApiError {
 		return new ApiError(StatusCodes.BAD_REQUEST, { errors: messages.flat() });
 	}
 
+	static notFound(message) {
```
Philip-Philippou1, line 13: Nice tidy up of the error code.

packages/applications-service-api/src/services/submission.service.js
```diff
@@ -66,7 +67,22 @@ const updateSubmission = async (submission) => {
...
+const getSubmission = (submissionId) => db.Submission.findOne({ where: { id: submissionId } });
+
+const completeSubmission = async (submissionId) => {
+	const submission = await getSubmission(submissionId);
+
+	if (!submission) {
+		throw ApiError.notFound(`Submission with ID ${submissionId} not found`);
+	}
+
+	console.log(`Found submission with ID ${submission.id}, email: ${submission.email}`);
```
Philip-Philippou1, line 79: I don't think we should log the email as it's personal information.

packages/applications-service-api/src/error/apiError.js
JamesP-KC: I removed this `commentsForPartyWithIDNotUpdated` function as it doesn't seem to be in use anywhere

packages/applications-service-api/src/services/submission.service.js
JamesP-KC: That's fair enough, I added this line just temporarily so the endpoint does "something" and was going to remove it in the next story. It's not really needed so I will remove now

Philip-Philippou1 (approved):
Looks good

PR #511, opened 2022-11-28
feat(submission-complete): create new submission complete page

Philip-Philippou1:
Added new page for submission complete.

- The submissionId has been added to the examinationSession data
- There is an assumption that project email and submissionId will be available.  Page will still be rendered without them as we don't want to fail the process at this stage.

### DIFF
--- a/packages/forms-web-app/__tests__/unit/controllers/examination/submission-complete/controller.test.js
+++ b/packages/forms-web-app/__tests__/unit/controllers/examination/submission-complete/controller.test.js
@@ -3,32 +3,34 @@ const {
 } = require('../../../../../src/controllers/examination/submission-complete/controller');
 
 const {
-	getExaminationSession
+	getExaminationSubmissionId
 } = require('../../../../../src/controllers/examination/session/examination-session');
+const {
+	getProjectEmailAddress
+} = require('../../../../../src/controllers/session/app-data-session');
 
 jest.mock('../../../../../src/controllers/examination/session/examination-session', () => ({
-	getExaminationSession: jest.fn()
+	getExaminationSubmissionId: jest.fn()
+}));
+
+jest.mock('../../../../../src/controllers/session/app-data-session', () => ({
+	getProjectEmailAddress: jest.fn()
 }));
 
 describe('examination/submission-complete/controller', () => {
 	describe('#getSubmissionComplete', () => {
-		const session = {
-			appData: {
-				ProjectEmailAddress: 'dummy.email@testing.gov.uk'
-			},
-			examination: {
-				submissionId: '1234'
-			}
-		};
+		const session = {};
 		const req = { session };
 		const res = {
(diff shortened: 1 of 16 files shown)


packages/forms-web-app/__tests__/unit/controllers/examination/submission-complete/controller.test.js
```diff
@@ -0,0 +1,58 @@
...
+			}
+		};
+		const req = { session };
+		const res = {
+			redirect: jest.fn(),
+			render: jest.fn(),
+			status: jest.fn(() => res)
+		};
+		describe('When getting the submission complete page', () => {
+			beforeEach(async () => {
```
jonathangoulding, line 30: no need for async ?

packages/forms-web-app/src/controllers/examination/process-submission/utils/process.js
```diff
@@ -20,6 +20,8 @@ const handleProcessSubmission = async (session) => {
 			}
 		}
 		logger.info('Processing files complete');
+		examinationSession.submissionId = submissionId;
```
jonathangoulding, line 23: should in a setter in the session dir. 

can be in the examination file.

jonathangoulding, line 23: You want the api call to /complete before setting the submission item.

it thats 2xx then continue otherwise throw. (Maybe a custom error / error log )

packages/forms-web-app/__tests__/unit/controllers/examination/submission-complete/controller.test.js
Philip-Philippou1: Good spot, removed

packages/forms-web-app/src/controllers/examination/process-submission/utils/process.js
Philip-Philippou1: Added a setter.

Philip-Philippou1: Added the complete API call.  You may want to take a look at the structure, there is a new exception which will log.

packages/forms-web-app/src/controllers/examination/submission-complete/controller.js
```diff
@@ -1,13 +1,29 @@
...
 
 const getSubmissionComplete = (req, res) => {
-	return res.render(submissionComplete.view);
+	try {
+		const { session } = req;
+		const examinationSession = getExaminationSession(session);
+
+		const pageData = {
+			submissionId: examinationSession.submissionId,
+			projectEmail: session.appData.ProjectEmailAddress
```
jonathangoulding, line 19: can be a getter ? in the top level session dir.

jonathangoulding (approved):

PR #516, opened 2022-11-30
fix(delete files on disk): add delete multiple file delete

jonathangoulding:


packages/forms-web-app/src/controllers/examination/file-upload/fileManagement.js
```diff
@@ -22,8 +22,16 @@ const deleteFileOnDisk = async (file) => {
 	await fs.unlinkSync(fileToDeletePath);
 };
 
+const iterateDeleteFileOnDisk = async (files) => {
+	for (const filesKey of files) {
+		const fileToDeletePath = fileUploadPath(filesKey.uniqueFileName);
+		await fs.unlinkSync(fileToDeletePath);
```
jonathangoulding, line 28: No real need to have this as sync other than to allow us to throw an error if the delete fails.

jonathangoulding (commented):
Mentioned use of sync

Philip-Philippou1 (approved):

PR #517, opened 2022-11-30
feat(applications-service-api): documents v3 endpoint

JamesP-KC:
- add initial implementation for `POST /v3/documents`
- remove implementation of old `POST /documents` (v1) endpoint. v2 kept for now, to be deleted after switch to v3.
- openapi updates 
  - mark `POST /documents` and `GET /v2/documents` as deprecated
  - add definition for `POST /v3/documents`


Philip-Philippou1 (approved):
Looks good, like the removal of V1.

PR #522, opened 2022-12-01
feat(applications-service-api): documents v3 endpoint apply search filters

JamesP-KC:
- implementation of search filtering for the `POST /v3/documents` endpoint

### DIFF
--- a/packages/applications-service-api/__tests__/unit/services/document.v3.service.test.js
+++ b/packages/applications-service-api/__tests__/unit/services/document.v3.service.test.js
@@ -18,12 +18,14 @@ describe('documentV3 service', () => {
 	beforeEach(() => jest.resetAllMocks());
 
 	describe('fetchDocuments', () => {
-		it('calls query api with correct params and returns document rows and total count', async () => {
+		beforeEach(() => {
 			mockFindAndCountAll.mockResolvedValueOnce({
 				count: 4,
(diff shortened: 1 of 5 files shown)


packages/applications-service-api/__tests__/unit/services/document.v3.service.test.js
```diff
@@ -43,15 +45,89 @@ describe('documentV3 service', () => {
...
+								{ category: "Developer's Application", filter_1: ['Plans', 'Reports'] },
+								{ stage: 1, filter_1: ['Deadline 2', 'Deadline 3'] }
+							]
+						}
+					]
+				}
+			});
+		});
+
+		it('calls query api with searchTerm if it is specified', async () => {
```
Philip-Philippou1, line 85: Tests look good, we could add a test for searchTerm combined and a Filter, but this will be picked up when we test API.

Philip-Philippou1 (dismissed):

packages/applications-service-api/__tests__/unit/services/document.v3.service.test.js
JamesP-KC: Yep that's a good scenario, I'll add a test for that

Philip-Philippou1 (approved):

PR #533, opened 2022-12-08
feat(applications-service-api): add filter label to document v3 response

JamesP-KC:
- exclude stage 0 from filters
- add label to document v3 filter response
- fix error when document filter type is empty array
- (openapi) make value and name required for /v3/documents request filter

### DIFF
--- a/packages/applications-service-api/src/utils/documentFilterLabelMapper.js
+++ b/packages/applications-service-api/src/utils/documentFilterLabelMapper.js
@@ -1,0 +1,24 @@
+const mapDocumentFilterLabel = (filterName, filterValue) => {
+	try {
+		return {
+			stage: {
+				1: 'Pre-application',
+				2: 'Acceptance',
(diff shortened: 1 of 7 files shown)


packages/applications-service-api/src/utils/documentFilterLabelMapper.js
```diff
@@ -0,0 +1,24 @@
...
+	try {
+		return {
+			stage: {
+				1: 'Pre-application',
+				2: 'Acceptance',
+				3: 'Pre-examination',
+				4: 'Examination',
+				5: 'Recommendation',
+				6: 'Decision',
+				7: 'Decided'
```
Philip-Philippou1, line 11: I think that may be Post-decision looking at the design - 
https://applications-prototype.up.railway.app/design-sprint-6/wide-nav/doc-library?filtersapplied=true

Philip-Philippou1 (approved):

PR #556, opened 2023-01-04
ASB-1012: Check Exam Timetable Date Before Allowing Access

christopher-semmens:


packages/forms-web-app/src/controllers/projects/examination-timetable.js
```diff
@@ -41,6 +41,8 @@ const eventSubmitButtonActive = (timetable) => {
 	);
 };
 
+const displayEvents = (dateOfNonAcceptance) => new Date(dateOfNonAcceptance) >= getDate();
```
Philip-Philippou1, line 44: Is this correct - we want to only view timetable when current date is after the dateOfNonAcceptance (timetable published date)

Philip-Philippou1 (approved):

PR #559, opened 2023-01-09
fix(examination-timetable.njk): remove links to project overview page

Philip-Philippou1:
Remove links to overview page from exam timetable page.
Also make the closed date footer logic only appear if the exam timetable is shown.


JamesP-KC (approved):
LGTM

PR #560, opened 2023-01-09
ASB-924: Submitting for change link

christopher-semmens:


packages/forms-web-app/__tests__/unit/controllers/examination/submitting-for/controller.test.js
```diff
@@ -0,0 +1,121 @@
...
+					expect(res.render).toHaveBeenCalledWith('error/unhandled-exception');
+				});
+			});
+		});
+	});
+	describe('#postSubmittingFor', () => {
+		describe('When handling a submitting page for post request', () => {
+			describe('and there is an error', () => {
+				const mockErrorValue = {
+					errors: { a: 'b' },
```
jonathangoulding, line 74: I think this should assert the error id is correct. 

This checking the errors keys length in the controller. 

A:B is less descriptive than using the coupled error id field. 

you are not asserting the correct id has been used. 

E.G:

```
			describe('and there errors with the body', () => {
				beforeEach(() => {
					req.body = {
						errors: { 'examination-add-another-deadline-item': 'an error' },
						errorSummary: [{ text: 'Error summary', href: '#' }]
					};
					getPageData.mockReturnValue({ text: 'mock page data' });
					postAddAnotherDeadlineItem(req, res);
				});
				it('should render the error page', () => {
					expect(res.render).toHaveBeenCalledWith('pages/examination/add-another-deadline-item', {
						errorSummary: [
							{
								href: '#',
								text: 'Error summary'
							}
						],
						errors: {
							'examination-add-another-deadline-item': 'an error'
						},
						text: 'mock page data'
					});
				});
			});
```

jonathangoulding: This asserts the id from the error should match the id filed used in the error check.

packages/forms-web-app/src/controllers/examination/submitting-for/utils/get-back-link-url.js
```diff
@@ -0,0 +1,25 @@
...
+	routesConfig: {
+		examination: {
+			directory,
+			pages: { applicant, checkYourAnswers, yourInterestedPartyNumber }
+		}
+	}
+} = require('../../../../routes/config');
+
+const getBackLinkUrl = (query, session) => {
+	if (isQueryModeEdit(query)) return `${directory}${checkYourAnswers.route}`;
```
jonathangoulding, line 15: single return using the redirect URL like do everywhere else ?

I see the point in throwing the error. If this was a case we would need to do this everywhere else ?

packages/forms-web-app/src/controllers/examination/submitting-for/utils/get-page-data.js
```diff
@@ -0,0 +1,21 @@
...
+const {
+	routesConfig: {
+		examination: {
+			pages: { submittingFor }
+		}
+	}
+} = require('../../../../routes/config');
+
+const getPageData = (query, session) => {
+	return {
```
jonathangoulding, line 12: unnecessary return ? 

can give the object straight back

packages/forms-web-app/__tests__/unit/controllers/examination/submitting-for/utils/get-redirect-url.test.js
```diff
@@ -0,0 +1,63 @@
...
+				let result;
+				beforeEach(() => {
+					isQueryModeEdit.mockReturnValue(true);
+					result = getRedirectUrl(mockQuery, '');
+				});
+				it('should return the check your answers URL', () => {
+					expect(result).toEqual('/examination/check-your-answers');
+				});
+			});
+			describe('and the selected submittin for value is myself', () => {
```
jonathangoulding, line 29: spelling on submitting

packages/forms-web-app/src/controllers/examination/submitting-for/utils/get-redirect-url.js
```diff
@@ -0,0 +1,26 @@
...
+	if (isQueryModeEdit(query)) return `${directory}${checkYourAnswers.route}`;
+
+	switch (selectedSubmittingForValue) {
+		case submittingFor.options[1].value:
+			return `${directory}${nameMyself.route}`;
+		case submittingFor.options[2].value:
+			return `${directory}${nameOrganisation.route}`;
+		case submittingFor.options[3].value:
+			return `${directory}${nameAgent.route}`;
+		default:
```
jonathangoulding, line 21: Throw an exception, since this is an exceptional case.

And throw it outside the switch, it would be more readable. Otherwise it sounds like "the default case is exceptional".


https://stackoverflow.com/questions/2567932/throwing-exception-vs-returning-null-value-with-switch-statement

jonathangoulding: Happy to keep it but want to ask the question.

jonathangoulding (approved):

PR #569, opened 2023-01-12
ASB-885: Move Examination Library to Top of Documents Page

christopher-semmens:
### Task
https://pins-ds.atlassian.net/jira/software/c/projects/ASB/boards/166?modal=detail&selectedIssue=ASB-885


packages/forms-web-app/src/controllers/projects/documents/utils/documents/get-examination-library-document-html.js
```diff
@@ -0,0 +1,13 @@
+const fileSizeDisplayHelper = require('../../../../../lib/file-size-display-helper');
+const fileTypeDisplayHelper = require('../../../../../lib/file-type-display-helper');
+
+const isValidDocument = (document) => document?.path && document?.mime && document?.size;
```
jonathangoulding, line 4: why is optional chaining used ?

packages/forms-web-app/src/controllers/projects/documents/utils/documents/documents-view-model.js
```diff
@@ -1,17 +1,21 @@
 const { formatDate } = require('../../../../../utils/date-utils');
+
+const formatDocumentToViewModel = (document) => ({
```
jonathangoulding, line 3: Good improvement but unnecessary to spend time on this and adding an extra thing to look at on the PR.

packages/forms-web-app/src/controllers/projects/documents/controller.js
```diff
@@ -21,13 +24,20 @@ const getApplicationDocuments = async (req, res) => {
 		const pageFeatureToggles = featureToggles();
 		const pageDataObj = await pageData(case_ref);
 
-		const { documents, filters, pagination } = await searchDocuments(case_ref, query);
+		const { documents, examinationLibraryDocument, filters, pagination } = await searchDocuments(
+			case_ref,
+			query
+		);
 		const documentsView = getDocuments(documents);
+		const examinationLibraryDocumentHtml = getExaminationLibraryDocumentHtml(
```
jonathangoulding, line 32: this should be returned by get documents and exported as two objects. 

This relates to documents so should come from there ?

packages/forms-web-app/src/controllers/projects/documents/utils/documents/get-examination-library-document-html.js
christopher-semmens: I removed this check in favour of a ternary condition applied that checks if there is a document before the function is called. The reason optional chaining was previously in place was because the function was called regardless if an examination library document was found or not.

packages/forms-web-app/src/controllers/projects/documents/controller.js
```diff
@@ -21,13 +21,15 @@ const getApplicationDocuments = async (req, res) => {
 		const pageFeatureToggles = featureToggles();
 		const pageDataObj = await pageData(case_ref);
 
-		const { documents, filters, pagination } = await searchDocuments(case_ref, query);
+		const { documents, examinationLibraryDocumentHtml, filters, pagination } =
```
jonathangoulding, line 24: This doesn't look right. 

This is returning data in raw form not transforming anything. 

By putting the html conversion in here it's now doing more than one thing. 

Plus you have added conditional logic into something that should be a simple call to get the data. 

get the examination response from here. 

And in get documents do the html conversions. 

This separation is important for extendability. We want one place to get the response (responses) and the subsequent functions to manipulate the shape of the data.

jonathangoulding: with the documents get you would add this extra documents to the document object for the view. 

`documents: { documentsView, examinationLibraryDocumentHtml}`'

this way anything to do with documents in the view is scoped to that key.

packages/forms-web-app/src/controllers/projects/documents/utils/documents/get-examination-library-document-html.js
jonathangoulding: cool good solution.

jonathangoulding (approved):

PR #572, opened 2023-01-16
feat(forms-web-app): redis cache config set up

JamesP-KC:
changes to the web app to enable using redis as session store:

- update version of redis node library
- amend local config to use redis server with a password, as the azure redis server does
- helper function to parse the redis connection string provided by azure redis

### DIFF
--- a/packages/forms-web-app/package.json
+++ b/packages/forms-web-app/package.json
@@ -52,7 +52,7 @@
 		"nunjucks": "^3.2.3",
 		"nunjucks-date-filter": "^0.1.1",
 		"pino": "^6.13.2",
-		"redis": "3.1.2",
+		"redis": "^4.5.1",
 		"slugify": "^1.6.5",
 		"string-collapse-white-space": "^9.1.0",
 		"string-strip-html": "^8.3.0",
--- a/packages/forms-web-app/src/lib/session.js
+++ b/packages/forms-web-app/src/lib/session.js
@@ -1,22 +1,59 @@
 const config = require('../config');
+const logger = require('./logger');
+const { createClient } = require('redis');
+const ConnectRedis = require('connect-redis');
 
-module.exports = () => {
-	const { sessionSecret } = config.server;
+const configureSessionStore = (session) => {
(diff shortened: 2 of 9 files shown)


packages/forms-web-app/package.json
```diff
@@ -52,7 +52,7 @@
 		"nunjucks": "^3.2.3",
 		"nunjucks-date-filter": "^0.1.1",
 		"pino": "^6.13.2",
-		"redis": "3.1.2",
+		"redis": "^4.5.1",
```
jonathangoulding, line 55: Do we want to allow it to update compatible version with the ^ ?

Appreciate we are not consistent across the repo and we don't have an npmrc with 'save-exact=true' inside.

packages/forms-web-app/src/lib/session.js
```diff
@@ -1,22 +1,55 @@
...
+			maxAge: 1000 * 60 * 60 * config.server.sessionLengthInHours
+		}
 	};
 
 	if (config.server.useSecureSessionCookie) {
-		sessionConfig.cookie.secure = true; // serve secure cookies
+		sessionConfig.cookie.secure = true;
+	}
+
+	if (config.featureFlag.useRedisSessionStore) {
```
jonathangoulding, line 24: Could this a be a func something like configureRedis ? (Just do it inside the file and leave the tests test)

It should help separate out Redis and the default session store. 

Something like - 
`if (config.featureFlag.useRedisSessionStore) sessionConfig.store = redisThing(session)`

Connect redis  `const RedisStore = ConnectRedis(session);` could be put inside the function aswell

packages/forms-web-app/package.json
jonathangoulding: Happy with that. Will resolve.

packages/forms-web-app/src/lib/session.js
JamesP-KC: Yeah that sounds good to me 👍

jonathangoulding (approved):

PR #578, opened 2023-01-18
buggered git

jonathangoulding:


packages/forms-web-app/src/controllers/examination/session/examination-session.js
```diff
@@ -50,6 +50,16 @@ const getExaminationSubmissionId = (session) => {
 
 const getExaminationEmailAddress = (session) => getExaminationSession(session).email;
 
+const getExaminationSubmissionRetryErrorCount = (session) => {
+	const examinationSession = getExaminationSession(session);
+	return examinationSession.retryErrorCount;
+};
+const setExaminationSubmissionRetryErrorCount = (session) => {
```
christopher-semmens, line 57: Just a question around the naming as it seems to be doing 2 things, creating or incrementing it. Maybe a chance to split it out?

packages/forms-web-app/src/controllers/examination/submission-error/utils/get-page-data.js
```diff
@@ -0,0 +1,17 @@
+const { getProjectEmailAddress } = require('../../../../controllers/session/app-data-session');
+
+const { getExaminationSubmitCommentsURL } = require('./get-examination-submit-comments-url');
+
+const getPageData = (session) => {
+	const pageData = {
+		id: 'examination-submission-error',
```
christopher-semmens, line 7: Does this make sense to go into the config to be consistent?

packages/forms-web-app/src/controllers/examination/session/examination-session.js
jonathangoulding: I think it's a setter. Consistent with other places we set defaults.

christopher-semmens (dismissed):

christopher-semmens (approved):

PR #580, opened 2023-01-23
feat(register personal information): refactor address

jonathangoulding:
refactor address into single controller and view


DavidBeale (approved):
Seems to make sense to me 👍

DavidBeale (approved):

PR #584, opened 2023-01-23
Fix: ASB-885 Always Display Examination Library Document

christopher-semmens:


packages/forms-web-app/src/controllers/projects/documents/utils/documents/search-examination-library-document.js
```diff
@@ -1,11 +1,13 @@
...
-const searchExaminationLibraryDocument = async (body) => {
-	const localBody = JSON.parse(JSON.stringify(body));
-	localBody.searchTerm = examinationLibraryDocumentSearchTerm;
-	const { data } = await searchDocumentsV3(localBody);
+const searchExaminationLibraryDocument = async (case_ref) => {
+	const examinationLibraryDocumentBody = getBody(case_ref, {
+		searchTerm: examinationLibraryDocumentSearchTerm
+	});
+	const { data } = await searchDocumentsV3(examinationLibraryDocumentBody);
 	return data?.documents[0];
```
jonathangoulding, line 11: can be more explicit if this does not exists than using optional chaining ? 

or should it just resort to undefined ?

jonathangoulding (approved):

PR #595, opened 2023-01-31
Task asb 1111 e2e scenario tests 1

akano-baba:
e2e happy-path user journey tests


JamesP-KC:
Might be an idea to add the `.vs` directory to the `.gitignore` file, as we wouldn't necessarily want IDE config in the repository

akano-baba:
image

JamesP-KC:
As part of moving the e2e tests into the monorepo, it would be good if you could put your new directory inside the `packages` directory. Perhaps name it `e2e-tests`, so the we keep the old tests in the root directory but the new ones live in a newly created `packages/e2e-tests` directory.
We can then  pick up amending the mono repo build process separately

JamesP-KC (changes requested):
A couple of changes I think it would be good to do:

1. add the `.vs` directory to the `.gitignore` file, as we wouldn't necessarily want IDE config in the repository so it should be ignored

2. As part of moving the e2e tests into the monorepo, it would be good if you could put your new directory inside the `packages` directory. Then we keep the old tests in the root directory (`./e2e-tests`) but the new ones live in a newly created `./packages/e2e-tests` directory. We can then pick up amending the mono repo build process separately as this will require dev input and outside the scope of this ticket.

Philip-Philippou1 (approved):

JamesP-KC (approved):
Changes look good.
I pushed a commit which removes the `.vs` directory as it was committed before we amended `.gitignore`, so it needed to be manually removed.

Going to approve on the basis the changes look reasonable; although I haven't run the e2e tests.

Minor nitpick, but I would prefer `packages/e2e_tests` uses a dash like the rest of the packages. So renaming it to `packages/e2e-tests` would be ideal, but not going to insist on it.

akano-baba:
> Changes look good. I pushed a commit which removes the `.vs` directory as it was committed before we amended `.gitignore`, so it needed to be manually removed.
> 
> Going to approve on the basis the changes look reasonable; although I haven't run the e2e tests.
> 
> Minor nitpick, but I would prefer `packages/e2e_tests` uses a dash like the rest of the packages. So renaming it to `packages/e2e-tests` would be ideal, but not going to insist on it.

I have renamed the folder to e2e-tests

PR #597, opened 2023-02-01
Fix/asb 845

RaySage:


RaySage:
Not sure if this need res.status(404).render('error/not-found');

DavidBeale (dismissed):
I guess the `res.status(404)` should be included

RaySage:
Thanks @DavidBeale - I've added the 404 status response and updated the test

DavidBeale (approved):

PR #613, opened 2023-02-08
fix(asb-430): correction to content for have your say guide

RaySage:


RaySage:
**Notes**

The correct content was hard coded in interested-party.njk, and the side bar version was getting different content from data in step-by-step-navigation.njk.

So anoher macro has been created, now we have one for main content and on for side bar, as there are styling and other differences.

Both macros now use the single source of data from step-by-step-navigation.njk, and this has been updated with the correct content from interested-party.njk.

Which macro used is controlled by an additional peoperty - templateID - if it's 'main' the macro for the main page is used, otherwise the side bar as default, this saves specifying the side macro in the other pages.

Maybe a different pattern could be used to include these re-useable page elements?

DavidBeale (dismissed):

christopher-semmens (commented):
@RaySage Branch name is not in the required format

RaySage:
Hi @christopher-semmens I have renamed the branch, sorry about that!

PR #617, opened 2023-02-09
ASB-1028: Mobile View - Show Filters on a New Page

christopher-semmens:
### Task
https://pins-ds.atlassian.net/jira/software/c/projects/ASB/boards/166?modal=detail&selectedIssue=ASB-1028


packages/forms-web-app/__tests__/unit/controllers/projects/documents/controller.component.test.js
```diff
@@ -184,7 +184,7 @@ describe('#getApplicationDocuments', () => {
 						}
 					],
 					examinationLibraryDocumentHtml:
-						'<p><a href="mock/path">View examination library (PDF, 225KB)</a> containing document reference numbers</p>',
+						'<p><a class="govuk-link" href="mock/path">View examination library (PDF, 225KB)</a> containing document reference numbers</p>',
```
DavidBeale, line 187: I wonder if we could use SASS to lift these classes' styles into the elements as default.

DavidBeale (approved):

packages/forms-web-app/__tests__/unit/controllers/projects/documents/controller.component.test.js
christopher-semmens: @DavidBeale I had the same idea and did implement this approach but recently took it out as there have been numerous situations when we have needed different styling on links which caused a lot of overriding properties.

DavidBeale (approved):

PR #625, opened 2023-02-14
Feature/asb 891 filter documents by date

akano-baba:
Added some date filter tests


akano-baba (commented):
Fixed broken tests

DavidBeale (approved):
Shame we are still using Cucumber.

Would be nice if step definitions where less specific and copy be reused.

PR #634, opened 2023-02-15
feat: further copy changes

DavidBeale:


christopher-semmens (commented):
Hi @DavidBeale Just to check, do these changes have any impact on the e2e tests we already have in place?

DavidBeale:
If they do, we should delete those tests

christopher-semmens (approved):

PR #638, opened 2023-02-20
Feature/asb 1076 pagination documents page

jonathangoulding:


packages/forms-web-app/src/views/components/ui/results-per-page.njk
```diff
@@ -0,0 +1,12 @@
+{% macro uiResultsPerPage(resultsPerPage) %}
+	<p >
+		<span class="govuk-!-font-weight-bold govuk-!-margin-right-3">Results per page</span>
```
christopher-semmens, line 3: In the past accessibility testing has picked us up on applying bold to tags suggesting to use `<strong>...</strong>` instead.

packages/forms-web-app/src/views/components/ui/results-per-page.njk
```diff
@@ -0,0 +1,12 @@
...
+		<a href="{{ resultsPerPage.twentyFive.link }}">{{ resultsPerPage.twentyFive.size }}</a>
```
christopher-semmens, line 4: Maybe good to add an aria-label on these links?

jonathangoulding: Screenshot 2023-02-20 at 15 17 19

jonathangoulding: added this is the voice over response

christopher-semmens (approved):

PR #646, opened 2023-02-23
fix(a11y): file upload button focus visible

jonathangoulding:


packages/forms-web-app/src/sass/components/ui/file-upload.scss
```diff
@@ -0,0 +1,10 @@
+#file-upload {
```
christopher-semmens, line 1: Might be good to give the elements more specific class names and maybe even put it into the component just to reduce risk of any CSS contamination.

packages/forms-web-app/src/views/pages/examination/select-file.njk
```diff
@@ -154,8 +154,8 @@
 				{% endif %}
 			>
 				const target = document.documentElement;
-				const dropZone = document.querySelector('.dropZone');
-				const fileInput = document.getElementById('file-upload')
+				const dropZone = document.querySelector(".dropZone");
+				const fileInput = document.getElementById("file-upload");
```
christopher-semmens, line 158: Quotes correcting to double? Is this right? I think mine are single.

jonathangoulding: passing linter ?

DavidBeale: not a JS file - so prob no linting 🤔

jonathangoulding: 'fixed'

packages/forms-web-app/src/sass/components/ui/file-upload.scss
```diff
@@ -0,0 +1,10 @@
+.document-file-upload {
+	&:focus {
+		+ .govuk-button {
```
christopher-semmens, line 3: This could become `.document-file-upload__label`

christopher-semmens: Apologies, I meant that the label element with the `.govuk-button` class could have another class added to the element which would be `.document-file-upload__label`. How the input was defined before was fine.

christopher-semmens (approved):

PR #651, opened 2023-02-27
Feature/asb 1193 api advice get by

DavidBeale:


packages/applications-service-api/src/controllers/advice.js
```diff
@@ -40,5 +40,30 @@ module.exports = {
...
+		} catch (e) {
+			if (e instanceof ApiError) {
+				logger.debug(e.message);
+				res.status(e.code).send({ code: e.code, errors: e.message.errors });
+				return;
+			}
+			logger.error(e.message);
+			res
+				.status(StatusCodes.INTERNAL_SERVER_ERROR)
+				.send(`Problem getting advice ${adviceID} \n ${e}`);
```
JamesP-KC, line 66: we might want to send a JSON object here, i think the web-app will try and parse this as json and fail. This might be the case on other controllers too so we can address this as a tech debt ticket if it's a big change

JamesP-KC (dismissed):

packages/applications-service-api/src/controllers/advice.js
DavidBeale: I think the webapp just looks at the 500 status message - but yes, makes sense to have it in JSON if we are bothering to send it. 👍

DavidBeale: Updated on these new endpoints

JamesP-KC (approved):

PR #681, opened 2023-03-15
Feature/section 51 meeting and from

jonathangoulding:


packages/forms-web-app/src/pages/projects/section-51/utils/adviceVM.js
```diff
@@ -0,0 +1,41 @@
+const getAdviceName = ({ organisation, firstName, lastName }) => {
```
DavidBeale, line 1: I'd prob say a ViewModel was not a util - but at the same level as controller etc

DavidBeale (approved):

packages/forms-web-app/src/pages/projects/section-51/utils/adviceVM.js
jonathangoulding: good shout. Have lifted it out

PR #685, opened 2023-03-16
feat(data-service): set up new package, mssql and prisma

JamesP-KC:
ASB-1227

- creates a new package `data-service`, to manage new applications database access
- sets up sql server locally with docker
- sets up prisma, to manage sql server migrations and ORM


packages/back-office-service/prisma/migrations/20230316101542_init/migration.sql
```diff
@@ -0,0 +1,74 @@
+BEGIN TRY
+
+BEGIN TRAN;
+
+-- CreateTable
+CREATE TABLE [dbo].[Project] (
+    [id] INT NOT NULL IDENTITY(1,1),
+    [caseId] INT NOT NULL,
+    [caseReference] NVARCHAR(1000),
```
DavidBeale, line 9: id, caseId and caseRef ?

packages/back-office-service/prisma/migrations/20230316101542_init/migration.sql
```diff
@@ -0,0 +1,74 @@
...
+    [easting] INT,
+    [northing] INT,
+    [welshLanguage] BIT,
+    [mapZoomLevel] NVARCHAR(1000),
+    [secretaryOfState] NVARCHAR(1000),
+    [dateProjectAppearsOnWebsite] NVARCHAR(1000),
+    [dateOfDCOAcceptance] NVARCHAR(1000),
+    [anticipatedDateOfSubmission] NVARCHAR(1000),
+    [anticipatedSubmissionDateNonSpecific] NVARCHAR(1000),
+    [dateOfDCOSubmission] NVARCHAR(1000),
```
DavidBeale, line 30: NVARCHAR(1000) for a date?

DavidBeale (approved):
I think back-office-service would be quite a confusing name - as this is not the back office

Maybe just service-layer

packages/back-office-service/prisma/migrations/20230316101542_init/migration.sql
JamesP-KC: Yeah, that is excessive.. I'll have a look at defining lengths in the Prisma schema file

JamesP-KC: We are getting them from Back Office as Strings, so was going to store as we get them. Could convert to a date type though if that would be better?

packages/back-office-service/prisma/migrations/20230316101542_init/migration.sql
JamesP-KC: `id` is possibly not needed but `caseId` and `caseReference` come from Back Office. There is some difference between those right @Philip-Philippou-KC ?

we could potentially use `caseId` as our primary key instead of having our own internal `id` field. I think in reality, we're not going to use `id`, we'll be querying by `caseReference` (or maybe `caseId`)

packages/back-office-service/prisma/migrations/20230316101542_init/migration.sql
DavidBeale: I would say so - assuming they are ISO 8601 etc, would be better for our DB to have date types

packages/back-office-service/prisma/migrations/20230316101542_init/migration.sql
DavidBeale: We would need our own synthetic id for joining etc. If caseId is the BO office synthetic id, then we should not store that (and really the BO should not leak this out in its messages)

Philip-Philippou1: The caseId is BO DB auto-increment Id.  I think their API accesses the resources in their endpoints using that Id (need to confirm that).  All of our existing endpoints in our API use the caseReference so we need that.  We may need all 3, id for our internal joins, integrations with BO may need caseId for linkage, caseRef to satisfy our own API requirements.

packages/back-office-service/prisma/schema.prisma
```diff
@@ -0,0 +1,66 @@
...
+  projectName                                 String?
+  projectDescription                          String?
+  publishStatus                               String?
+  sector                                      String?
+  projectType                                 String?
+  sourceSystem                                String
+  stage                                       String?
+  projectLocation                             String?
+  projectEmailAddress                         String?
+  regions                                     String?
```
JamesP-KC, line 26: this field (`regions`) is actually an array of strings from Back Office.

Primsa has a JSON type, so we can store it as a stringified array,
or do our own normalisation an build up a table of regions and use
joins.

i am leaning toward json. Any thoughts @Philip-Philippou1 @DavidBeale

DavidBeale: I think the expectation would be to use a table. Else we would have been allowed to use a Document DB

JamesP-KC: i think the ask was to use sql server rather than nosql rather than avoid any features in sql server that are nosql inspired :) 

i think if we have a known list of regions, we could make this a table. i will check with back office

Philip-Philippou1 (approved):

PR #713, opened 2023-04-25
feat(applications-service-api): fetch application data sourced from Back Office

JamesP-KC:
- rework `api/v1/applications` endpoint to fetch from MSSQL db (Back Office data) if the `caseReference` is in a whitelist
    - retains the existing API contract

refactoring:
- move all the prisma functionality from `data-service` and into `applications-service-api` (and delete `data-service`)
- added integration test (`packages/applications-service-api/__tests__/integration/applications.test.js`)
- moved `async-route` helper from `forms-web-app` to `common`
- make `caseReference` the primary key in the Project table


packages/applications-service-api/src/lib/config.js
```diff
@@ -9,6 +9,13 @@
 const path = require('path');
 
 module.exports = {
+	backOfficeIntegration: {
+		applications: {
+			getApplication: {
+				caseReferences: ['EN0110004', 'EN0310004']
```
JamesP-KC, line 15: should the caserefs be in an env var so we can set different caserefs per environment?
e.g. `caseReferences: process.env.BACK_OFFICE_INTEGRATION_CASE_REFERENCES`

Philip-Philippou1: I think that would be much better, perhaps we could use that env var throughout then as we add functionality.

Philip-Philippou1 (approved):

packages/applications-service-api/src/lib/config.js
JamesP-KC: added in an env var. Value is a string (comma separated)

PR #730, opened 2023-06-15
ASB-1466: back office docs api

JamesP-KC:
- amends `POST /v3/documents` endpoint to fetch documents + filters from Applications DB (SQL Server) 
    - fetches from applications DB if caseref is included in config (`BACK_OFFICE_INTEGRATION_GET_DOCUMENTS_CASE_REFERENCES`)
    - otherwise, fetches from NI (mysql) DB as usual
    - router calls different controller function depending on whether caseref is BO or NI. 
    - separate 'service', 'repository' and 'mapper' functions for NI and BO data


packages/applications-service-api/__tests__/unit/utils/document.mapper.test.js
```diff
@@ -0,0 +1,193 @@
...
+	mapDocuments,
+	mapBackOfficeDocuments
+} = require('../../../src/utils/document.mapper');
+
+describe('document mapper functions', () => {
+	describe('mapDocuments', () => {
+		it('maps NI documents to v3 response body format', () => {
+			const niDocuments = [
+				{
+					id: 1000,
```
Philip-Philippou1, line 12: Is it worth adding some of the fields that are specific to NI to highlight the differences?
eg,
category
personal_name

packages/applications-service-api/src/repositories/document.backoffice.repository.js
```diff
@@ -0,0 +1,76 @@
...
+		]
+	};
+
+	if (query.searchTerm) {
+		whereClause['AND'].push({
+			OR: [
+				{ description: { contains: query.searchTerm } },
+				{ author: { contains: query.searchTerm } },
+				{ representative: { contains: query.searchTerm } },
+				{ mime: { contains: query.searchTerm } }
```
Philip-Philippou1, line 34: I never realised we search on mime, but you are correct we currently do, seems really odd, wonder if path would be more useful.

packages/applications-service-api/src/routes/documents.v3.js
```diff
@@ -2,9 +2,22 @@ const express = require('express');
...
+const { asyncRoute } = require('@pins/common/src/utils/async-route');
 
 const router = express.Router();
 
-router.post('', validateRequestWithOpenAPI, documentsV3Controller.getDocuments);
+const getDocumentsRoute = (req, res, next) => {
+	const backOfficeCaseReferences =
+		config.backOfficeIntegration.documents.getDocuments.caseReferences || [];
+
+	const route = backOfficeCaseReferences.includes(req.body.caseReference)
```
Philip-Philippou1, line 14: In the future we might want to extend this to have a global switch to BO, probably best on another ticket to handle all routes.

packages/applications-service-api/src/utils/document.mapper.js
```diff
@@ -0,0 +1,131 @@
...
+		filter1: document.filter1,
+		filter2: document.filter2,
+		description: document.description,
+		size: document.size,
+		mime: document.mime,
+		path: document.documentURI,
+		// status: 'Published', // should always be Published, so not needed in API response?
+		datePublished: document.datePublished,
+		// deadlineDate: "", // no equivalent in BO schema
+		// personalName: '', // no equivalent in BO schema
```
Philip-Philippou1, line 75: I think personalName is the one field I'd like us to understand a bit more.  The FE uses this at the moment mainly as a description type field.  Not much we can do as it isn't passed by BO so it is hopefully covered by the other fields (description).

JamesP-KC: From looking at the test data (would be good to see prod data for comparison), `personalName` looks to be much like `author`, which seems to be null on all the test data I've seen.

We have `author` in BO, so would suggest switching the frontend to use that instead of `personalName`.

packages/applications-service-api/src/routes/documents.v3.js
JamesP-KC: Yes, probably one to cover off in another ticket when we have a better idea how we want to go live with a BO/NI hybrid solution

packages/applications-service-api/src/repositories/document.backoffice.repository.js
JamesP-KC: Yeah I carried this over from the original /v1/documents endpoint, although in retrospect it doesn't seem very useful to search on `mime`, unless we have a usecase for searching by file type (word, excel, pdf, etc). Even then that need would probably be better served by a filter.

Happy to drop it, as the more fields you search across, the fuzzier the results become.

packages/applications-service-api/__tests__/unit/utils/document.mapper.test.js
JamesP-KC: Good call 👍

Philip-Philippou1 (approved):

PR #743, opened 2023-07-17
feat(applications-service-api): delete subscription endpoint

JamesP-KC:
adds `DELETE /api/v1/subscriptions/{caseReference}` endpoint

takes encrypted email as query param and publishes message to back office 'delete nsip subscription' topic, containing email + case ref

### DIFF
--- a/packages/applications-service-api/src/controllers/subscriptions.js
+++ b/packages/applications-service-api/src/controllers/subscriptions.js
@@ -3,7 +3,10 @@ const { sendSubscriptionCreateNotification } = require('../lib/notify');
 const { getApplication } = require('../services/application.v2.service');
 const ApiError = require('../error/apiError');
 const moment = require('moment');
-const { publishNSIPSubscription } = require('../services/backoffice.publish.service');
+const {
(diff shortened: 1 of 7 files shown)


packages/applications-service-api/src/controllers/subscriptions.js
```diff
@@ -46,14 +49,28 @@ const confirmSubscription = async (req, res) => {
...
 
-	res.status(200).send();
+	res.send();
+};
+
+const deleteSubscription = async (req, res) => {
+	const { caseReference } = req.params;
+	const { email } = req.query;
+
+	const project = await getApplication(caseReference);
```
Philip-Philippou1, line 61: Not a biggy but is it worth moving into it's own function?

Philip-Philippou1 (approved):

packages/applications-service-api/src/controllers/subscriptions.js
JamesP-KC: Thanks, have added a function and used it for both endpoints

PR #749, opened 2023-07-21
ASB-1615: add project update table

JamesP-KC:
- adds `ProjectUpdate` table to Applications database, with some seed data

other chore/refactoring:
- amend docker-compose config so mssql container runs by default
- create prisma seed script for some basic Project data
- update readme with info about managing prisma/sql server


packages/applications-service-api/prisma/schema.prisma
```diff
@@ -104,3 +104,18 @@ model Document {
 
   @@index([caseRef])
 }
+
+model ProjectUpdate {
+  id                   Int       @unique @default(autoincrement())
+  projectUpdateId      Int       @unique
+  caseReference        String
+  updateDate           String?
```
Philip-Philippou1, line 112: Looking at BO this is date the update is published.  We are looking to display the latest published update and our API will likely sort the response by this field.  I wonder if we should be enforcing it as mandatory.  Perhaps we can note it and do it later.

Philip-Philippou1 (dismissed):

packages/applications-service-api/prisma/schema.prisma
JamesP-KC: Confirmed it will be present when we receive the message. Will make it non-optional in our schema here

Philip-Philippou1 (approved):

PR #759, opened 2023-08-01
feat(relevant reps): date check to show rr or not

jonathangoulding:


packages/forms-web-app/src/pages/projects/relevant-representations/representations.js
```diff
@@ -70,7 +71,8 @@ exports.getRepresentations = async (req, res) => {
 		pageOptions,
 		searchTerm,
 		queryUrl,
-		commentsTypeFilterItems
+		commentsTypeFilterItems,
+		showReps: isDateAfterTodaysDate(applicationResponse.DateRRepAppearOnWebsite)
```
Philip-Philippou1, line 75: I think we may have to handle the case where the date is not set.  I think .DateRRepAppearOnWebsite  gets set just before they publish the reps so can this handle null or not set?

jonathangoulding: date not set will result in an invalid date and be false.

jonathangoulding: i'll add a test for the case.

Philip-Philippou1 (approved):

PR #760, opened 2023-08-01
feat(applications-service-api): add GET projectUpdates endpoint

JamesP-KC:
new endpoint to fetch project updates

`GET /api/v1/projectUpdates/{{caseReference}}`


Philip-Philippou1 (approved):
This looks good.  The one question I have is how we would handle updates to an existing update.  If BO publish an update will we upsert?  Just wondering if this endpoint could end up returning all versions of the same update?

JamesP-KC:
> This looks good. The one question I have is how we would handle updates to an existing update. If BO publish an update will we upsert? Just wondering if this endpoint could end up returning all versions of the same update?

Yeah we will upsert

PR #772, opened 2023-08-07
Feature/project info add document type endpoint

jonathangoulding:


packages/applications-service-api/src/routes/documents.v3.js
```diff
@@ -7,6 +7,12 @@ const { asyncRoute } = require('@pins/common/src/utils/async-route');
 
 const router = express.Router();
 
+router.get(
+	'/:caseReference',
+	validateRequestWithOpenAPI,
```
Yakitrak, line 12: Suggestion / question for discussion - Is it worth doing custom validation on controller instead of the swagger validation for the `caseReference`? For example if we are sent `Rule 8 Letter` instead of `Rule 8 letter` it will return a 404 but we can easily capitalise the `caseReference` param and compare it a capitalised enum of "RULE 8 LETTER". This will ensure a small capitalisation issue isn't going to cause any issues. What do you guys think?

packages/applications-service-api/__tests__/unit/controllers/documents.v3.test.js
```diff
@@ -183,4 +193,116 @@ describe('documentsV3 controller', () => {
 			});
 		});
 	});
+
+	describe('getDocumentByCaseReference', () => {
+		describe('NI', () => {
+			describe('dict test', () => {
```
Yakitrak, line 199: We can change the desc to something like "when document type is in wrong letter case"

packages/applications-service-api/__tests__/unit/controllers/documents.v3.test.js
```diff
@@ -183,4 +193,116 @@ describe('documentsV3 controller', () => {
...
+				expect(fetchNIDocumentsByType).toHaveBeenCalledWith({
+					caseReference: 'EN000001',
+					type: 'Rule 8 letter - notification of timetable for the examination'
+				});
+
+				expect(res._getStatusCode()).toEqual(StatusCodes.OK);
+				expect(res._getData()).toEqual({ data: 'mock document' });
+			});
+		});
+		describe('BO', () => {
```
Yakitrak, line 261: same comment as above

Yakitrak (changes requested):

packages/applications-service-api/api/openapi.yaml
```diff
@@ -301,6 +301,42 @@ paths:
...
+              'DCO DECISION LETTER (SOS)(APPROVE)',
+              'DCO DECISION LETTER (SOS)(REFUSE)'
+            ]
+      responses:
+        '200':
+          description: Returns list of documents
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/DocumentListWrapper'
```
JamesP-KC, line 336: `DocumentListWrapper` has all the pagination and filtering stuff, and this endpoint is returning just a single document so i think `DocumentV3` is what we want here

packages/applications-service-api/src/controllers/documents.v3.js
```diff
@@ -41,7 +47,59 @@ const buildFilters = (req) => ({
...
+	'RULE 6 LETTER': {
+		bo: 'Rule 6 letter',
+		ni: 'Rule 6 letter - Notification of the preliminary meeting and matters to be discussed'
+	},
+	'RULE 8 LETTER': {
+		bo: 'Rule 8 letter',
+		ni: 'Rule 8 letter - notification of timetable for the examination'
+	},
+	'EXAMINATION LIBRARY': {
+		bo: 'Examination library',
```
JamesP-KC, line 60: i think in back office the value is `Exam library` from what i've seen in the database. Might be worth checking in their repo, api spec or message schema

packages/applications-service-api/src/controllers/documents.v3.js
```diff
@@ -41,7 +47,59 @@ const buildFilters = (req) => ({
 	datePublishedTo: req.body.datePublishedTo
 });
 
+const documentTypeDict = {
```
JamesP-KC, line 50: it's a style/nitpick so feel free to ignore but i was thinking it might be tidier to delegate this mapping of `type` to the 'service' functions, and keep that DB specific stuff in the service or repository?

i was thinking pass the `type` as-is to `fetchBackOfficeDocumentsByType` / `fetchNIDocumentsByType`,
and then in those functions map to the generic value to its DB equivalent

JamesP-KC (changes requested):

packages/applications-service-api/src/middleware/validator/validateGetDocumentWithTypeRequestOpenAPI.js
```diff
@@ -0,0 +1,29 @@
+const { createRequestValidator } = require('../../utils/openapi');
+const ApiError = require('../../error/apiError');
+
+const validateGetDocumentWithTypeRequestOpenAPI = (req, res, next) => {
```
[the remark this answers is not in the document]
jonathangoulding: This was intentionally done so the raw query would be used in the controller. This should avoid any potential mistakes / bugs of thinking why is it all caps etc. 

happy to change it but we thought this could lead to confusion / potential bugs in the future.

packages/applications-service-api/src/controllers/documents.v3.js
jonathangoulding: where would this be stored ? 

i can not find a ref to it.

JamesP-KC: on back office this gets set when editing the metadata for a document - 

Screenshot 2023-08-10 at 15 49 29


possible values are defined here:
https://github.com/Planning-Inspectorate/back-office/blob/7399139990d1f745d190bef012d8977f0506e7bd/apps/web/src/server/applications/case/documentation-metadata/documentation-metadata.controller.js#L78-L99

so looks to me like its `Exam library`

and also those "decision letter" documents _are_ using the same naming as NI (albeit with slightly different casing)

jonathangoulding: fixed exam lib

JamesP-KC (approved):

PR #774, opened 2023-08-10
Feature/asb 1622 project location

christopher-semmens:
### Task
https://pins-ds.atlassian.net/browse/ASB-1622?atlOrigin=eyJpIjoiN2MzNDcwNmQ3YzFmNDQ2Mjk5NjA0OTViMzdjMjQ4MTMiLCJwIjoiaiJ9


JamesP-KC (dismissed):
Looks good to me. Just a couple of things around dotenv

Can we add a `.env.development` file, as an example of what the real `.env` file should look like? Something like:

```
OS_MAPS_API_KEY="value"
OS_MAPS_API_SECRET="value"
```

and we should also add `.env` to the [.gitignore](https://github.com/Planning-Inspectorate/applications-service/blob/main/.gitignore) file so it doesn't get accidentally committed.

JamesP-KC (approved):

PR #786, opened 2023-08-24
Fix/Exam timetable closing a date early

Yakitrak:
Ticket: https://pins-ds.atlassian.net/browse/ASB-1648

The issue was happening because some data had no time set on `DateTimeExaminationEnds` which meant it was treated as `00:00:00` therefore when checking on the closing day, it would always be considered closed instead of opened. This ticket creates a new function to change any non-set times to be considered `23:59:59` whilst still allowing other timestamps to work as expected.


packages/forms-web-app/__tests__/unit/utils/is-before-or-after-date.test.js
```diff
@@ -51,6 +52,27 @@ describe('#utils/is-before-or-after-date', () => {
...
+		describe('When given datetime is today 00:00:00', () => {
+			describe('and the current datetime is today 12:00:00', () => {
+				it('should return true', () => {
+					jest.useFakeTimers().setSystemTime(new Date('2020-01-01 12:00:00'));
+					const result = isBeforeTodayUTC('2020-01-01 00:00:00');
+					expect(result).toEqual(true);
+				});
+			});
+		});
+	});
```
JamesP-KC, line 75: maybe add a test case for a string with no time?

Yakitrak: Ah I did actually have that but removed it because the test cases was just duplicating the tests we have for the `isBeforeNowUTC` function. But I can add it if you think its worth having here too?

Yakitrak: Added on commit: https://github.com/Planning-Inspectorate/applications-service/pull/786/commits/2df44c0dfa13a4b22110838d8a41bf336b8d8cad

JamesP-KC: yeah worth having i think, as adding a time where it is missing is the main thing ` isBeforeTodayUTC` is doing

Yakitrak: https://github.com/Planning-Inspectorate/applications-service/pull/786/commits/dfa0f356a7d3e7d579aae63aea8b952ea892a5b7

JamesP-KC (approved):

PR #788, opened 2023-08-25
Feature/application endpoint pagination

Yakitrak:
Adds pagination functionality to the get applications endpoint.

Ticket: https://pins-ds.atlassian.net/browse/ASB-1739

### DIFF
--- a/packages/applications-service-api/__tests__/integration/applications.test.js
+++ b/packages/applications-service-api/__tests__/integration/applications.test.js
@@ -1,7 +1,9 @@
-const { APPLICATION_DB } = require('../__data__/application');
+const { APPLICATION_DB, APPLICATION_FO } = require('../__data__/application');
 const { request } = require('../__data__/supertest');
 
 const mockFindUnique = jest.fn();
+const mockFindAll = jest.fn();
+const mockCount = jest.fn();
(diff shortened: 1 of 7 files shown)


christopher-semmens (approved):

packages/applications-service-api/__tests__/integration/applications.test.js
```diff
@@ -10,65 +12,95 @@ jest.mock('../../src/lib/prisma', () => ({
...
+				expect(response.body).toEqual({
+					code: 400,
+					errors: ['\'caseReference\' must match pattern "^[A-Za-z]{2}\\d{6,8}$"']
+				});
 			});
 		});
+	});
+	describe('get all applications', () => {
+		it('happy path', async () => {
+			mockFindAll.mockResolvedValueOnce([{ dataValues: APPLICATION_DB }]);
```
JamesP-KC, line 86: the project data in `APPLICATION_DB` is for a Back Office case, so it might be good to use some test data for a NI case instead. For example, `projectData` in the `__tests__/unit/controllers/applications.test.js` file

the fields used in the Back Office and NI databases are a bit different, so our API response will need to be consistent, so there will be some mapping to do - for example in the NI cases, the field `Stage` is an int, and from this API we want to return `stage` as an string

PR #833, opened 2023-09-22
refactor(applications-service-api): application service interface

JamesP-KC:
Refactoring of the applications services to provide a more generic interface to fetch an application, without exposing database implementation detail.

Introduces new function `getApplication` (`services/application.service.js`). This function calls either `getBackOfficeApplication` or `getNIApplication` depending on which system the application belongs to, and returns the application data in a common/generalised format.

This PR stops short of changing the rest of the code to use this new interface. Over time, we should continue to refactor and replace calls to `getBackOfficeApplication` or `getNIApplication` with the `getApplication` function, eventually making those two functions non-exported/private.

Other changes:
- New mapper functions in `utils/application.mapper.js` (`mapNIApplicationToApi` and `mapBackOfficeApplicationToApi`) to map data from database format (NI or Back Office, respectively) into a more generic "API" format for use internally - and eventually to return from the `/applications/{caseReference}` endpoint. Changing the API response will be a breaking change for the frontend, so is beyond the scope of this PR.
- Rename `services/application.service` (as it was), to `services/application.ni.service`. `services/application.service` now contains the new generic `getApplication` function and the `getBackOfficeApplication` function. Eventually, `application.ni.service` can be deleted once NI is decommissioned, leaving just `application.service`.
- Moving some util functions into their own file and adding tests: `utils/mapLocation.js`
- Reorganising NI/BO/API Application test data, moving it into `__tests__/__data__/application.js`, adding some more realistic dates (to the best of my knowledge)


Philip-Philippou1 (approved):

Philip-Philippou1 (approved):
⁠Looks good to me, it's a bit tricky to review as there are are now 3 application services but the approach is great.  If we need to change the strategy for determining if it's a back office case it can be done in 1 place the service.  

This is going to need some regression!

PR #835, opened 2023-09-25
fix: add secondary sort when sorting stage in get applications

Yakitrak:
Ticket: https://pins-ds.atlassian.net/browse/ASB-1877


packages/applications-service-api/__tests__/unit/services/application.ni.service.test.js
```diff
@@ -152,6 +151,20 @@ describe('application.ni.service', () => {
 							order: [[sort, 'ASC']]
 						});
 					});
+
+					it('when sort key is stage, calls findAll with secondary sort projectName', async () => {
+						// Act
```
JamesP-KC, line 156: are the comments an auto-generated thing? i guess they might be useful for larger tests with a lot of mocks, but with the majority of tests being only a few lines of code i don't find they add much as jest mocking and expectation syntax is pretty readable already. If you find them useful then fair enough but if it's autogenerated i would lean toward removing them

JamesP-KC (approved):

packages/applications-service-api/__tests__/unit/services/application.ni.service.test.js
Yakitrak: Its not auto generated, I like the pattern as it separates everything to be more readable. But fair enough on small tests its doesn't add much, I do it out of habit. In the future I'll only keep these comments for the bigger tests.

Useful link: https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/

PR #837, opened 2023-09-25
feat(applications-service-api): allow subscribing to project updates for ni application

JamesP-KC:
- rework `controllers/subscriptions.js` to use new `getApplication` service function, which allows the endpoint to process `caseReference`s that are either for applications in the NI database, or from Back Office
- refactoring:
  - make `getBackOfficeApplication` private as this was the only place it was in use and is now superseded by `getApplication`
  - move `addMapZoomLevelAndLongLat` to `application.mapper.js` and simplify implementation
  - fix incorrect default zoom level


Philip-Philippou1 (approved):
LGTM

PR #850, opened 2023-10-03
feat(applications-service-api): submission back office integration

JamesP-KC:
## Describe your changes

ASB-1818

Reworking `POST /api/v1/submissions/{caseReference}` endpoint to allow submissions for both NI and Back Office cases.

Details about a Submission for a Back Office case (the "metadata") is published to the Back Office message queue. If the Submission contains a user uploaded file, that will be uploaded to Blob Storage (out of scope of this PR). If the Submission contains a written representation ("comment"), then a PDF file containing the comment is generated and will be uploaded to Blob Storage (ASB-1830).

Refactoring (3d6acf4b2868212ebb1817599a2449340b326330):
- Introduces general `submission.service` which contains new logic for submitting Back Office submissions, and all existing NI related functions have been moved to a new `submission.ni.service`
- Changes to the API schema:
  - Response body now contains only `submissionId`, which is all frontend is using
  - The `submissionId` property (in request and response) can be string or number. NI uses int, BO will be string.
- Moved NI Submission database functions to `submission.ni.repository`

## Type of change 🧩

- [ ] Documentation update


packages/applications-service-api/src/services/submission.service.js
```diff
@@ -1,105 +1,41 @@
...
+	isBackOfficeSubmission(submission.metadata.caseReference)
+		? createBackOfficeSubmission(submission)
+		: createNISubmission(submission);
+
+const completeSubmission = async (submissionId) => completeNISubmission(submissionId);
+
+const createBackOfficeSubmission = async (submission) => {
+	const { metadata } = submission;
+
+	if (!metadata.submissionId) metadata.submissionId = uuid.v4();
```
Philip-Philippou1, line 17: We might want to check this is ok, as currently users get:
Your reference number
12569

If these need to be communicated over the phone a uuid might be awkward.

Appreciate it's not easy without a DB.

JamesP-KC: Yeah that's fair, UUID in this case is a bit overkill. I've pushed up a different approach based on case reference and timestamp. That should give enough uniqueness considering the volume of submissions

Yakitrak (commented):
LGTM but wondering if we can / should add integrations tests for this (`packages/applications-service-api/__tests__/integration`).

JamesP-KC:
> LGTM but wondering if we can / should add integrations tests for this (`packages/applications-service-api/__tests__/integration`).

Yeah that's a good shout. I've added some to cover the back office submission journey (excluding file upload, will add that in next PR)

packages/applications-service-api/__tests__/integration/submission.test.js
```diff
@@ -0,0 +1,149 @@
...
+				interestedPartyReference: '1234567879',
+				lateSubmission: false,
+				name: 'Joe Bloggs',
+				sensitiveData: false,
+				submissionType: 'Comments on LIRs'
+			};
+
+			describe('submission with written representation', () => {
+				describe('request with submissionId provided', () => {
+					it('uploads pdf of representation, publishes message, and returns submissionId', async () => {
```
Yakitrak, line 51: This might be partly preference but I think it would be better for the tests to be isolated (not have "and" in test blocks) - have the `sendMessage` assertion in a different block and the returns `submissionId` in its own one. It makes it clear what is failing if something starts breaking. You can just throw the `submissionRequest` in a beforeEach.

Yakitrak (approved):
LGTM! Left a minor comment about the tests but feel free to ignore

packages/applications-service-api/__tests__/integration/submission.test.js
JamesP-KC: It's a good shout, and happy to go with a different convention. I usually write the tests quite "fat" with lots of expectations, mainly to reduce repetition of the arrange/act parts but as you say - that could be handled in a `beforeEach`. They certainly read better when you don't have lots of text in the `it` and break it down into a more nested structure. I'll be updating these tests in my next ticket so will do some restructuring in the next PR

PR #868, opened 2023-10-11
feat(applications-service-api): upload submission file to blob storage

JamesP-KC:
## Describe your changes

ASB-1830

Builds upon https://github.com/Planning-Inspectorate/applications-service/pull/850 to add functionality for uploading file to Azure Blob Storage container, and publishing the location/path in the message to the message queue.

Also restructured the `integration/submission.test.js` test to have smaller, more specific assertions in each `it`.

## Type of change 🧩

- [ ] Documentation update


docker-compose.yml
```diff
@@ -20,6 +20,8 @@ services:
     container_name: applications-service-api
     environment:
       APPLICATIONS_WEB_BASE_URL: http://forms-web-app:9004
+      BACK_OFFICE_BLOB_STORAGE_DEADLINE_SUBMISSION_URL: http://localhost:10000/
```
Yakitrak, line 23: Might be a silly question, where is this running locally or it is just a placeholder?

packages/applications-service-api/src/lib/blobStorage.js
```diff
@@ -0,0 +1,18 @@
+const { DefaultAzureCredential } = require('@azure/identity');
+const { BlobServiceClient } = require('@azure/storage-blob');
+
+const config = require('./config');
+
+const upload = async (buffer, mimeType, path) => {
+	const client = new BlobServiceClient(
```
Philip-Philippou1, line 7: Just thinking about the unhappy path where that client can't be created (missing/incorrect config), would we get a reasonable error logged?

Philip-Philippou1 (approved):

docker-compose.yml
JamesP-KC: Good question! This is the default address for Blob Storage when you run [Azurite](https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio)

packages/applications-service-api/src/lib/blobStorage.js
JamesP-KC: I tested it out and the error is a bit vague in the scenario where the storage container URL is undefined, so will add in some error handling and logging

JamesP-KC: Added a couple of try/catch blocks to allow us to log a more meaningful error. The exception details will get caught and logged higher up the stack so haven't logged the error itself here.

docker-compose.yml
Yakitrak: Ah I see. I asked because if its going to be running as a separate docker container we could make use of the docker network configs to reference the url.

Will making it work locally end to end be part of the next ticket?

Philip-Philippou1 (approved):

PR #872, opened 2023-10-18
Fix/asb 1963 submission filename

JamesP-KC:
## Describe your changes

ASB-1968, ASB-1818

Updates to the `POST /api/v1/submissions/{submissionId}` and `POST /api/v1/submissions/{submissionId}/complete` endpoints

- aadb0c2d9bc8e012608937bdacf25f90b97fa4b2 simplify the file naming conventions used when uploading files to blob storage container. The PDF file generated for a written rep will have the name `Written-Representation.pdf` (previously had the submissionID included in the file name). A user uploaded file will be uploaded as-is and the file name will not be changed. 
- c53c62916d601274f98ff7c247ca4260d4c6630f fix issue with api not returning a 400 error if the `caseReference` property was missing from the request body but the `email` _was_ included. Updated the error message to be: `"must include both 'email' and 'caseReference'"`.

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/packages/applications-service-api/src/controllers/submissions.js
+++ b/packages/applications-service-api/src/controllers/submissions.js
@@ -1,6 +1,7 @@
 const { StatusCodes } = require('http-status-codes');
 const { createSubmission, completeSubmission } = require('../services/submission.service');
 const { getDate } = require('../utils/date-utils');
+const ApiError = require('../error/apiError');
 
 const createSubmissionController = async (req, res) => {
 	const submissionRequestData = buildSubmissionData(req);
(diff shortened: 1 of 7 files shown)


packages/applications-service-api/src/controllers/submissions.js
```diff
@@ -10,11 +11,21 @@ const createSubmissionController = async (req, res) => {
...
 const completeSubmissionController = async (req, res) => {
-	await completeSubmission({
+	const submissionDetails = {
 		submissionId: req.params.submissionId,
 		caseReference: req.body?.caseReference,
 		email: req.body?.email
-	});
+	};
+
+	if (
```
Philip-Philippou1, line 20: If there is no body or empty body then I think the existing Front end call for NI integration would get past this check and work.  Is that by design?

I think we may need a front end change to send the body but perhaps that is not for this ticket.

JamesP-KC: Yeah, that's intentional. The existing NI call has no body so I've made email and caseref optional (but if you provide one, then you must provide the other)

Philip-Philippou1 (approved):

Philip-Philippou1 (approved):

PR #884, opened 2023-11-01
feat: exam-timetable frontend use BO integration 

Yakitrak:
## Describe your changes

Ticket: https://pins-ds.atlassian.net/browse/ASB-1895

This PR confirms NI (regression) and back office case references work on the exam timetable journeys. There is minor updates to seeding and mapper to match back office to existing NI. There is also a bug fix to ensure we only show the "Have your say" button on a deadline if there is existing deadline items.

For regression tests locally (ensuring previous functionality works, use any case id which has appropriate data except  `BC0110001`

For back-office integration testing locally, use `BC0110001` as thats been seeded with the data and is in the list for the feature flag `PROJECT_MIGRATION_CASE_REFERENCES`. 

For testing on dev/test, data will need to be seeded into the DB so the API and frontend can use it as needed. 

## Type of change 🧩

- [ ] Documentation update


packages/forms-web-app/src/utils/timetables/check-timetable-state.js
```diff
@@ -13,6 +13,8 @@ const hasTimetableDeadlineStarted = (dateTimeDeadlineStart) =>
 	setTimeToStartOfDay(dateTimeDeadlineStart) <= getDateNow() ||
 	isNullSQLDate(new Date(dateTimeDeadlineStart));
 
+const hasDeadlineItemsList = (description) => description?.includes('* ');
```
JamesP-KC, line 16: is there an existing function somewhere that parses `description` that we could use? 
i'm assuming `description` gets turned into an array of line items somewhere else in the code (i did a quick search but couldn't find anything so may be wrong).

Yakitrak: image

Its converted to markdown here in `event-view-model.js`, but I felt it probably makes more sense to do this check. within the `check-timetable-state.js` - as thats where we will use it. Or do you mean we should use the same lib `marked` and check if it contains `ul` instead of checking for `*`?

JamesP-KC: ahh right, so it converts the markdown directly into html.. In my head it was going from markdown to json (simple array of strings for the line items), then to html.. my reservation was that trying to check the `description` manually might be error prone (like getting a rogue `*` in there somewhere that isn't a list item). But happy to approve ,we can always revisit if we get some unexpected descriptions coming through that break things

JamesP-KC (approved):

packages/forms-web-app/src/utils/timetables/check-timetable-state.js
Yakitrak: Yeah, I had reservations about a rogue `*` as well, for example bolding also uses a `*` but hopefully specifically checking for a `*` followed by a space should mitigate this. With a space, it won't bold so a `* ` should only be for lists. But yeah like you said, if anything happens I can pick it up for a better fix.

PR #886, opened 2023-11-01
feat: add app insights sdk and feature toggle

JamesP-KC:
## Describe your changes

ASB-1971

- Add feature flag for enabling use of Application Insights
- Initialise Application Insights library on app startup if feature flag is enabled

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/packages/applications-service-api/src/main.js
+++ b/packages/applications-service-api/src/main.js
@@ -1,9 +1,19 @@
 /* istanbul ignore file */
 
+const appInsights = require('applicationinsights');
+const config = require('./lib/config');
 const logger = require('./lib/logger');
 const server = require('./server');
 
 async function main() {
+	if (config.featureFlag.useApplicationInsights) {
+		try {
+			appInsights.setup().start();
+		} catch (err) {
+			logger.warn({ err }, 'Application insights failed to start: ');
+		}
+	}
+
 	server();
 }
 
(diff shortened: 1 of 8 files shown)


packages/applications-service-api/src/main.js
```diff
@@ -1,9 +1,19 @@
...
 
+const appInsights = require('applicationinsights');
+const config = require('./lib/config');
 const logger = require('./lib/logger');
 const server = require('./server');
 
 async function main() {
+	if (config.featureFlag.useApplicationInsights) {
+		try {
+			appInsights.setup().start();
```
Yakitrak, line 11: I am not familiar with azure insights but taking a look at the docs - do we not need to pass a connection string when doing `.setup()`?

Philip-Philippou1: Looks like appeals have activated with the following:

		appInsights.setup().setAutoDependencyCorrelation(true, true).setSendLiveMetrics(true);
		appInsights.defaultClient.context.tags[appInsights.defaultClient.context.keys.cloudRole] =
			'appeals-service-api';
		appInsights.start();
		
	I think the setAutoDependencyCorrelation is true by default, 
	the setSendLiveMetrics we may need to send the metrics to Azure,
	the  defaultClient.context.tags may be around splitting the insight by web app and api which may be useful.

JamesP-KC: The library should pick up the env var automatically

https://github.com/microsoft/ApplicationInsights-node.js#getting-started
> Configure the local SDK by calling appInsights.setup('YOUR_CONNECTION_STRING');, using the connection string you grabbed in step 2. Or put it in the APPLICATIONINSIGHTS_CONNECTION_STRING environment variable and call appInsights.setup() without parameters.

Yakitrak (approved):

PR #918, opened 2023-11-21
feat: back office integration extend get s51 advice by case reference endpoint 

Yakitrak:
## Describe your changes

Ticket: https://pins-ds.atlassian.net/browse/ASB-2026

PR extends `/api/v1/advice?caseRef={caseRef}` for back office integration for specific case references using the `BACK_OFFICE_INTEGRATION_GET_ADVICE_CASE_REFERENCES` flag. Also includes refactor to follow the repository-service design pattern. 

```
http://localhost:3000/api/v1/advice?caseRef=BC0110001
```

```
http://localhost:3000/api/v1/advice?caseRef=TR040009
```

## Type of change 🧩

- [ ] Documentation update


packages/applications-service-api/src/controllers/advice.js
```diff
@@ -2,78 +2,83 @@ const { StatusCodes } = require('http-status-codes');
...
-			if (e instanceof ApiError) {
-				logger.debug(e.message);
-				res.status(e.code).send({ code: e.code, errors: e.message.errors });
-				return;
-			}
-			logger.error(e.message);
-			res.status(StatusCodes.INTERNAL_SERVER_ERROR).send({
-				message: `Problem getting advice \n ${e}`
-			});
+		if (totalItems === 0) {
```
JamesP-KC, line 24: Do we want this behaviour of throwing an error if the caseref is valid but there's no advice? This `adviceNotFound` error message seems more for the scenario where you look up Advice by ID and a record with the given ID isn't found. 

Unless we have an AC for changing the behaviour to return 404, then returning 200 with an empty array is more in line with other endpoints

packages/applications-service-api/src/controllers/advice.js
```diff
@@ -2,78 +2,83 @@ const { StatusCodes } = require('http-status-codes');
...
-		try {
-			logger.debug(`Retrieving advice by ID...`);
+		res.status(StatusCodes.OK).send({
+			advice: advice,
+			totalItems: totalItems,
+			itemsPerPage: itemsPerPage,
+			totalPages: totalPages,
+			currentPage: currentPage
+		});
+	} catch (e) {
```
JamesP-KC, line 35: I think we can update this endpoint to use our newer pattern of handling errors at the router level. 

You can remove the `catch` and just `throw` errors in the controller, if you add the `asyncRoute` wrapper function in the router.

I did this for another old endpoint here: https://github.com/Planning-Inspectorate/applications-service/pull/917/files#diff-5fd35a5e8220a0d300bf5a4af2f79b81b9c7e91bb41cdb48e65fcce5725fac10R8

JamesP-KC (commented):
Looks good, couple of comments on the controller

packages/applications-service-api/src/controllers/advice.js
Yakitrak: Good point, I wasn't too sure, I did this to follow what the timetable was doing (which was updated by me recently but had this beforehand too). I don't know where we would throw `adviceNotFound`, would we be searching to see if the case ref exists in the project table to confirm that the case exists but we found no advice (this sounds may be over complicated)?

JamesP-KC: I think it makes sense for the `adviceNotFound` error to remain in the `getAdviceById` function, so it returns 404 if the given `id` for `GET /advice/{id}` is not found in Advice table.

For the `GET /advice` endpoint, I think it makes sense to return 400 when caseRef is missing (as you already are), but I wouldn't go as far as validating that caseRef exists in the Projects table. So if we remove the `if (totalItems === 0) { throw ApiError.adviceNotFound(caseRef)` condition, I think that covers all the ACs without regression or introducing new behaviour

packages/applications-service-api/api/openapi.yaml
```diff
@@ -936,7 +936,7 @@ paths:
         - in: query
           name: caseRef
```
Philip-Philippou1, line 937: I missed this while defining and refining but would it make sense to call this query param caseReference.
There are a lot of inconsistencies in the API so would be good to converge to a consistent convention.

packages/applications-service-api/src/utils/advice.mapper.js
```diff
@@ -0,0 +1,26 @@
+const mapBackOfficeAdviceToApi = (backOfficeAdvice) => {
+	return backOfficeAdvice.map((advice) => {
+		const [firstName, ...lastName] = advice && advice.from ? advice.from.split(' ') : undefined;
+		return {
+			s51Enquiry: true,
```
JamesP-KC, line 5: I think this should be named `section51Enquiry`, as per existing API / NI db

https://github.com/Planning-Inspectorate/applications-service/blob/3f3b942c396129dfcbe884807b13c5f44d48cdd2/packages/applications-service-api/src/models/advice.js#L24

packages/applications-service-api/__tests__/__data__/advice.js
```diff
@@ -0,0 +1,71 @@
...
+		enquiryMethod: 'Email',
+		industrySector: 'Energy',
+		caseReference: 'EN010009',
+		firstName: 'Joe',
+		lastName: 'Bloggs',
+		organisation: 'The organisation',
+		enquiryDetail: 'Do we need more energy',
+		adviceGiven: 'Yes we do',
+		respondedBy: 'Joe Bloggs',
+		section1Enquiry: true,
```
JamesP-KC, line 55:
```suggestion
		section51Enquiry: true,
```

packages/applications-service-api/api/openapi.yaml
Yakitrak: Thanks Philip, will do this in an additional PR as discussed.

JamesP-KC (approved):
Couple more minor comments on naming but looks good to me. Nice work 👍

PR #928, opened 2023-11-29
chore(schema.prisma): Update Representations table schema

Philip-Philippou1:
Representations schema to change based on Service User changes

## Type of change 🧩

- [ ] Documentation update
- [x ] Other (please explain in the description section above)


Yakitrak (commented):
James knows much more about this so will leave the approval to him but had a question - why can't we do both migrations in one go so instead of committing two folders in one PR, we can do it in one?

packages/applications-service-api/prisma/migrations/20231129122831_remove_rep_redacted_fd/migration.sql
```diff
@@ -0,0 +1,25 @@
...
+
+  - You are about to drop the column `redacted` on the `Representation` table. All the data in the column will be lost.
+
+*/
+BEGIN TRY
+
+BEGIN TRAN;
+
+-- AlterTable
+ALTER TABLE [dbo].[Representation] DROP COLUMN [redacted];
```
JamesP-KC, line 12: You could combine the two migrations into one. The idea being to take this `ALTER TABLE` line and put it after the `DROP CONSTRAINT` in the other migration

packages/applications-service-api/prisma/schema.prisma
```diff
@@ -61,40 +61,13 @@ model Representation {
   referenceId                    String?
   status                         String?
   dateReceived                   DateTime?
-  originalRepresentation         String    @db.NVarChar(4000)
-  redacted                       Boolean   @default(false)
-  redactedRepresentation         String?   @db.NVarChar(4000)
-  redactedBy                     String?
-  redactedNotes                  String?   @db.NVarChar(4000)
+  representationComment          String?   @db.NText
```
JamesP-KC, line 64: `NText` is deprecated which is why it was `NVarChar(4000)` before. Back Office are using `NText` so we can go for consistency (and both have to switch when it is fully deprecated), or use `NVarChar` which is the [recommended replacement](https://learn.microsoft.com/en-us/sql/t-sql/data-types/ntext-text-and-image-transact-sql?view=sql-server-ver16) and preempt it

packages/applications-service-api/prisma/migrations/20231129122831_remove_rep_redacted_fd/migration.sql
Philip-Philippou1: Good shout and much better merging into 1.

packages/applications-service-api/prisma/schema.prisma
Philip-Philippou1: Originally my concern was that BO could have a 4100 char rep and we would truncate.  Thought I'd make it consistent now then we can align on length in future.

JamesP-KC (approved):

PR #948, opened 2023-12-12
feat: extend get representation by id endpoint for bo integration

Yakitrak:
## Describe your changes

Ticket: https://pins-ds.atlassian.net/browse/ASB-1996

PR extends `/api/v1/representatiin/{representationId}?caseReference={caseReference}` for back office integration for specific case references using the `BACK_OFFICE_INTEGRATION_GET_REPRESENTATION_CASE_REFERENCES` flag. It includes:
- Adding and enforcing a required query parameter `caseReference` - this is required so we can route BO case references via feature flag correctly
- Frontend is updated to send the new parameter
- Refactor to follow the service repository design pattern
- Added seeding

```
http://localhost:3000/api/v1/representations/10?caseReference=BC0110001
```

```
http://localhost:3000/api/v1/representations/10?caseReference=EN010009
```

## Type of change 🧩

- [ ] Documentation update


packages/applications-service-api/src/repositories/representation.backoffice.repository.js
```diff
@@ -0,0 +1,13 @@
+const { prismaClient } = require('../lib/prisma');
+
+const getRepresentationById = async (representationId) => {
+	return prismaClient.representation.findUnique({
+		where: {
+			representationId
+		}
```
JamesP-KC, line 7: We could fetch the `represented` and `representative` details here as part of this query:
```
{
  include: {
    represented: true,
    representative: true
  }
}
```

We would need to add a relation in the schema:
```
representative         ServiceUser?               @relation("representative", fields: [representativeId], references: [serviceUserId], onDelete: NoAction, onUpdate: NoAction)
  represented            ServiceUser?               @relation("represented", fields: [representedId], references: [serviceUserId], onDelete: NoAction, onUpdate: NoAction)
```

It does end up doing separate queries rather than JOINs so the benefit is more about leveraging Prisma rather than implementing the extra service user repository to do the same thing

packages/applications-service-api/__tests__/unit/utils/representation.mapper.test.js
```diff
@@ -0,0 +1,23 @@
+const { mapBackOfficeRepresentationToApi } = require('../../../src/utils/representation.mapper');
+const {
+	REPRESENTATION_BACKOFFICE_RESPONSE,
+	REPRESENTATION_BACKOFFICE_DATA
+} = require('../../__data__/representation');
+const { SERVICE_USERS_BACKOFFICE_DATA } = require('../../__data__/serviceUser');
+const { BACK_OFFICE_DB_DOCUMENTS } = require('../../__data__/documents');
+describe('representation.mapper', () => {
+	describe('mapBackOfficeRepresentationToApi', () => {
```
JamesP-KC, line 9: a test for the case where `representativeId` / the `representative` param is null would be good, as that will be a common scenario

packages/applications-service-api/__tests__/integration/representations.test.js
```diff
@@ -17,21 +29,110 @@ jest.mock('../../src/models', () => ({
...
-			// Arrange
-			mockFindOne.mockResolvedValue(REPRESENTATION_NI_DB[0]);
-			// Act
-			const response = await request.get('/api/v1/representations/2');
-			// Assert
-			expect(mockFindOne).toBeCalledWith({ where: { ID: '2' }, raw: true });
-			expect(response.status).toEqual(200);
-			expect(response.body).toEqual(REPRESENTATION_NI_DB[0]);
+	config.backOfficeIntegration.representations.getRepresentations.caseReferences = ['BC0110001'];
+	describe(' GET /api/v1/representations/{representationId}?caseReference={caseReference}', () => {
```
JamesP-KC, line 52: a test case for a rep without representative would be good here too

packages/applications-service-api/src/repositories/representation.backoffice.repository.js
Yakitrak: Yeah we considered having an actual relation between them in the schema but we can't guarantee what order the messages will come in from BO so if get a represented (service user) first but would not have the representation. I believe @Philip-Philippou-KC also mentioned another issue with having a relation in prisma on this.

Philip-Philippou1: The ordering is the big one, we need to make message processing idempotent, if there is no service user there we will be constraints errors.

JamesP-KC: As discussed - foreign key constraints prevent this from being feasible due to the `Representation` row depending on the `ServiceUser` row being inserted first, which we can't guarantee due to messages from the queue possibly arriving out of order.

JamesP-KC (approved):

PR #953, opened 2023-12-15
feat: extend get all representations for case reference endpoint for bo integration

Yakitrak:
## Describe your changes

Ticket: https://pins-ds.atlassian.net/browse/ASB-1996

PR extends `/api/v1/representation?caseReference={caseReference}` for back office integration for specific case references using the `BACK_OFFICE_INTEGRATION_GET_REPRESENTATION_CASE_REFERENCES` flag.

The seeding will be done in the next ticket: https://pins-ds.atlassian.net/browse/ASB-2097 

```
http://localhost:3000/api/v1/representations?caseReference=BC0110001
```

```
http://localhost:3000/api/v1/representations?caseReference=EN010009
```

## Type of change 🧩

- [ ] Documentation update


packages/applications-service-api/prisma/schema.prisma
```diff
@@ -54,29 +54,31 @@ model Document {
...
   status                String?
   dateReceived          DateTime?
-  representationComment String?   @db.NText
+  representationComment String?      @db.NText
   representationFrom    String?
   representationType    String?
   registerFor           String?
   representedId         String?
   representativeId      String?
+  representative        ServiceUser? @relation("representative", fields: [representativeId], references: [serviceUserId], onDelete: NoAction, onUpdate: NoAction)
```
Philip-Philippou1, line 70: I don't think we want to add these relationships, they are covered with Id's abbove.

packages/applications-service-api/prisma/schema.prisma
```diff
@@ -86,8 +88,10 @@ model ServiceUser {
   organisationName String?
   role             String?
   serviceUserType  String?
-  createdAt        DateTime? @default(now())
-  modifiedAt       DateTime? @default(now())
+  representative   Representation[] @relation("representative")
```
Philip-Philippou1, line 91: Think this is going to give us key constraint issues, don't think we want these?

packages/applications-service-api/src/services/representation.service.js
```diff
@@ -42,20 +46,57 @@ const createQueryFilters = (query) => {
...
+	const representationsData = await Promise.all(
+		representations.map(async (representation) => {
+			if (!representation.representedId) return;
+			const represented = await getServiceUserByIdBORepository(representation.representedId);
+			if (!represented) return;
+			const representative = await getServiceUserByIdBORepository(representation.representativeId);
+			return { representation, represented, representative };
+		})
+	);
+	const representationsDataWithRepresented = representationsData.filter((r) => r?.represented);
```
Philip-Philippou1, line 68: Not sure if this filter is necessary, I think the block above enforces a represented to exist.

packages/applications-service-api/prisma/schema.prisma
Yakitrak: I thought I had reverted this! Thanks for catching this.

packages/applications-service-api/prisma/schema.prisma
Yakitrak: I thought I had reverted this! Thanks for catching this.

packages/applications-service-api/src/services/representation.service.js
Yakitrak: The block above is a map function so it will return null but the null represented  element will still exist in the array. Then this line will remove that from the array so we don't return it in the response. 

More explanation: The reason I used the map above like this was to do promise.all and speed up concurrent requests.

Philip-Philippou1 (approved):

PR #995, opened 2024-01-26
ASB-2161 - Reduce Exam timetable API calls from middleware

christopher-semmens:
## Describe your changes

https://pins-ds.atlassian.net/browse/ASB-2161?atlOrigin=eyJpIjoiOGE2ZTM1YzRlZmVmNDVmYmI3YzMxODBhYWUzYzJhY2IiLCJwIjoiaiJ9

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/packages/forms-web-app/src/utils/timetables/get-timetables-state.js
+++ b/packages/forms-web-app/src/utils/timetables/get-timetables-state.js
@@ -1,4 +1,8 @@
 const { getTimetables } = require('../../lib/application-api-wrapper');
+const {
+	getTimetablesSession,
+	setTimetablesSession
+} = require('../../pages/projects/examination-timetable/_session/timetables');
 const { isTimetableDateOfEventPast } = require('./check-timetable-state');
 const { isTimetableTypeOfEventDeadlineOpen } = require('./check-timetable-state');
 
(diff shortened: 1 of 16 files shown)


packages/forms-web-app/src/utils/timetables/get-timetables-state.js
```diff
@@ -13,19 +17,30 @@ const getOpenEventDeadlineTimetables = (timetables) =>
...
-		response = openEventDeadlineTimetables.length > 0;
+const getAllTimetables = async (session, case_ref) => {
+	let allTimetables = getTimetablesSession(session, case_ref);
+
+	if (!allTimetables) {
+		const { data } = await getTimetables(case_ref);
+
+		allTimetables = data?.timetables || [];
+
+		setTimetablesSession(session, case_ref, allTimetables);
```
Philip-Philippou1, line 28: I think this may lead to an unwanted side effect.  
If I'm understanding this we will be storing the timetable for every user for every case they visit.  We may run the risk of running out of storage in redis if there is a spike.

I may be wrong, but I think this introduces the storage of the timetable data in session much earlier.

christopher-semmens: I see your point. We could have an issue when users visit multiple projects which would cause all the timetables to be added to a project case reference key. We could end up with a large amount of data.

The reasoning for saving the data to the caseRef was to avoid having to create more complexity in the logic for detecting whether or not we have already checked the project. This way, if the check is undefined we know we need to do another api call.

As an alternative, we could only save a boolean value against each project, which would mean a much smaller object. Let me know you can foresee any potential issues with that method.

Philip-Philippou1: I really like the simple boolean method if possible.  
In most cases current implementation would be fine but we do introduce some risk and we are affectively creating a cache per user.  If there is a really big spike in users, it's going to be quite a lot of replicated and stored data.

Philip-Philippou1 (approved):
LGTM

PR #1021, opened 2024-02-07
feat: add 204 response to api root path

emil-placheta:
Task: https://pins-ds.atlassian.net/browse/ASB-2231

Add 204 to api root 

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/packages/applications-service-api/src/routes/index.js
+++ b/packages/applications-service-api/src/routes/index.js
@@ -15,6 +15,8 @@ const submissionRouter = require('./submissions');
 const subscriptionRoute = require('./subscriptions');
 const adviceRoute = require('./advice');
 
+router.get('/', (req, res) => res.sendStatus(204));
+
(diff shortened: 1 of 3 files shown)


packages/applications-service-api/src/routes/index.js
```diff
@@ -15,6 +15,8 @@ const submissionRouter = require('./submissions');
 const subscriptionRoute = require('./subscriptions');
 const adviceRoute = require('./advice');
 
+router.get('/', (req, res) => res.sendStatus(204));
```
emil-placheta, line 18: should this be router.all() instead?

Yakitrak: router.get is better since it only matches whereas router.all would match everything (post, delete etc).

Yakitrak (changes requested):
LGTM to me, will approve after the open api changes we discussed on call

Yakitrak (approved):

PR #1037, opened 2024-02-20
fix(api): fix advice mapper incorrectly mapping 'from' column

emil-placheta:
## Describe your changes
Mapper incorrectly mapping 'from' column

Ticket: https://pins-ds.atlassian.net/browse/ASB-2294

## Type of change 🧩

- [ ] Documentation update


packages/applications-service-api/src/utils/advice.mapper.js
```diff
@@ -17,7 +17,8 @@ const mapBackOfficeAdviceToApi = (advice) => {
 };
 
 const mapCommonAdviceFieldsToApi = (advice) => {
-	const [firstName, ...lastName] = advice && advice.from ? advice.from.split(' ') : undefined;
+	const [firstName, ...lastName] = advice && advice.agent ? advice.agent.split(' ') : undefined;
```
Philip-Philippou1, line 20: Have we moved the problem to advice.agent ?  agent can be empty, user can create s51 advice with from being only name, does this work?

Looks like it should but want to make sure.

emil-placheta: Ah, I assumed that `agent` (where we get first and last name) was required and always available and only `from` (organisation) was optional, but you are right, it's _either_ first name + lastname _OR_ organisation thats required (or both as org can be optional).

This change will return empty strings if we can't get the name out of `agent` and FE will default to from (org). If somehow neither are present, FE shows `Anonymous`

Philip-Philippou1 (approved):

PR #1069, opened 2024-03-06
feat: update register to have your say project link

cprieststephens:
## Describe your changes

2311: https://pins-ds.atlassian.net/jira/software/c/projects/ASB/boards/166?selectedIssue=ASB-2311
Update the project link on the register to have your say confirmation page.

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/packages/forms-web-app/src/pages/projects/register/_common/complete/controller.js
+++ b/packages/forms-web-app/src/pages/projects/register/_common/complete/controller.js
@@ -1,4 +1,4 @@
-const { getProjectsURL } = require('../../../../projects/_utils/get-projects-url');
+const { getProjectsIndexURL } = require('../../../index/_utils/get-projects-index-url');
 const { getKeyFromUrl } = require('../../../../../controllers/register/common/get-key-from-url');
 const logger = require('../../../../../lib/logger');
 const { viewModel } = require('./_utils/viewModel');
@@ -9,8 +9,8 @@ const {
 
 const getRegisterCompleteController = (req, res) => {
 	try {
-		const { session } = req;
-		const { caseRef } = session;
+		const { params, session } = req;
+		const { case_ref: caseRef } = params;
 		const key = getKeyFromUrl(req.originalUrl);
 		const { ipRefNo } = getSessionBase(session, key);
 		const { email } = getSession(session, key);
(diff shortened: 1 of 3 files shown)


packages/forms-web-app/src/pages/projects/register/_common/complete/controller.js
```diff
@@ -18,7 +19,7 @@ const getRegisterCompleteController = (req, res) => {
 			...viewModel[key],
 			ipRefNo,
 			email,
-			nsipProjectLink: nsipProjectLink(req.session.appData)
```
christopher-semmens, line 21: Having a quick glance of the codebase - it looks like this is the only place the 'nsipProjectLink' function is being called. Might be a good opportunity to raise a tech debt ticket to investigate if it is redundant and if so, to remove it.

packages/forms-web-app/src/pages/projects/register/_common/complete/controller.js
```diff
@@ -10,6 +10,7 @@ const {
 const getRegisterCompleteController = (req, res) => {
 	try {
 		const { session } = req;
+		const { caseRef } = session;
```
christopher-semmens, line 13: I would be inclined to retrieve the case reference from the params. If we had a case where another controller did not have the caseRef added to the locals we would need to retrieve it from the params so following that rule should help to keep the codebase more consistent.

```suggestion
		const { params } = req;
		const { case_ref } = params;
```

packages/forms-web-app/src/pages/projects/register/_common/complete/controller.js
```diff
@@ -18,7 +19,7 @@ const getRegisterCompleteController = (req, res) => {
 			...viewModel[key],
 			ipRefNo,
 			email,
-			nsipProjectLink: nsipProjectLink(req.session.appData)
+			projectURL: getProjectsURL(caseRef)
```
christopher-semmens, line 22: This should use the getProjectsIndexURL function.

```suggestion
			projectsIndexURL: getProjectsIndexURL(caseRef)
```

packages/forms-web-app/src/pages/projects/register/_common/complete/controller.js
cprieststephens: Thanks, will do.

christopher-semmens (approved):

PR #1134, opened 2024-04-22
feat: validation migration function

Yakitrak:
## Describe your changes

Ticket: https://pins-ds.atlassian.net/browse/ASB-2428

Adds a validate migration endpoint called via function app endpoint. It is called with a `caseReference`, and returns all entities (application, timetable, advice, representation, documents) differences for the application. The response omits any values which are identical and returns only the differences. It will traverse through nested objects and arrays.

Note: Existing endpoint services are used so any mandatory filters such as excluding draft documents will apply here.

JSON Diff library: https://www.npmjs.com/package/json-diff

### Dev endpoint: 
https://pins-app-applications-service-applications-api-dev-ukw-001.azurewebsites.net/validate-migration/BC0110001

### Function endpoint: 
Endpoint can be retrieved from Azure - not listed here as contains auth code.
Call with POST request with body:
```
{
    "caseReference": "BC0110001"
}

```
image

## Type of change 🧩

- [ ] Documentation update


packages/applications-service-api/src/repositories/document.ni.repository.js
```diff
@@ -14,7 +14,7 @@ const fetchDocuments = async (requestQuery) => {
 	const dbQuery = {
 		where,
 		order: [['date_published', 'DESC'], ['id']],
-		offset: (requestQuery.page - 1) * requestQuery.itemsPerPage,
+		offset: (requestQuery.page - 1) * requestQuery.itemsPerPage || 0,
```
Philip-Philippou1, line 17: Is this because we normally default the itemsPerPage in the FE and now it may not be sent when calling from the function?

Philip-Philippou1 (approved):

packages/applications-service-api/src/repositories/document.ni.repository.js
Yakitrak: Yes, otherwise sequalize threw an error because we were passing undefined (on NaN - can't recall which). But this wouldn't happen from the usual endpoints as they always have defaults.

PR #1204, opened 2024-06-25
fix: bypass ni sequelize connection on training env

Yakitrak:
## Describe your changes

Ticket: https://pins-ds.atlassian.net/browse/APPLICS-479

PR uses a mock sequelize connection if using training environment (determined by BACK_OFFICE_INTEGRATION_GET_APPLICATIONS  feature flag being set as 'BO'), which would bypass the NI connection errors that occurred because training environment cannot access NI. This approach was chosen over a memory DB as it requires no new dependancies or more memory, whilst still ensuring everything is built correctly

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/packages/applications-service-api/__tests__/unit/controllers/representations.test.js
+++ b/packages/applications-service-api/__tests__/unit/controllers/representations.test.js
@@ -86,6 +86,7 @@ const returnData = {
 };
 
 jest.mock('../../../src/lib/config.js', () => ({
+	...jest.requireActual('../../../src/lib/config.js'),
 	logger: {
 		level: process.env.LOGGER_LEVEL || 'info'
 	}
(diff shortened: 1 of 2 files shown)


beejjacobs (approved):
Seems a good method, that way as environments move over to BO only they won't try and connect to NI either.

I might suggest pulling out of some the logic into a small util function, that way the sequelize instance could be returned and be `const` here. Given this code could be removed once NI is retired though, not worth doing!

Yakitrak:
> Seems a good method, that way as environments move over to BO only they won't try and connect to NI either.
> 
> I might suggest pulling out of some the logic into a small util function, that way the sequelize instance could be returned and be `const` here. Given this code could be removed once NI is retired though, not worth doing!

Thanks! And yeah I considered splitting it out as well but because we will likely remove it soon, I kept the change as small as possible.

PR #1270, opened 2024-07-29
Added new common volume package to docker-compose to correct build issue on Windows machines

johnmoscrip:
## Added new common volume package to docker-compose to correct build issue on Windows machines

Update to docker-compose.yml to include new file reference

Prior to this change the application service solution build would fail with a "Cannot find module '@pins/common/src/utils/redis'" error when attempting to do a "npm run dev".  Including a reference to the module in docker-compose corrected this error.  The fix has been confirmed by myself and Ben Jacobs.  There is nothing to test on the front end for this fix.  There is no ticket for this issue.

## Type of change 🧩

- [X] Bug fix (non-breaking change which fixes an issue)

- [X] I have performed a self-review of my own code
- [X] I have double checked this work does not include any hardcoded secrets or passwords

### DIFF
--- a/packages/forms-web-app/src/lib/application-api-wrapper.js
+++ b/packages/forms-web-app/src/lib/application-api-wrapper.js
@@ -178,10 +178,10 @@ exports.wrappedSearchDocumentsV3 = async (body) => {
 	});
 };
 
-exports.wrappedPostSubmissionComplete = async (submissionId) => {
+exports.wrappedPostSubmissionComplete = async (submissionId, body) => {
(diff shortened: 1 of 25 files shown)


Yakitrak (changes requested):
Looks out of date from main as its got 25 file changes (instead of just the docker file).

johnmoscrip:
> Looks out of date from main as its got 25 file changes (instead of just the docker file).

Yep, that's odd.  Let me check.

PR #1284, opened 2024-08-05
chore: change favicon crown to updated crown icon

BenSurgisonPINS:
## Describe your changes

APPLICS-226 - https://pins-ds.atlassian.net/browse/APPLICS-226

This will require anyone developing locally to re-run the build prior to running the containers in docker.

- Removed all images from the public images folder for the Queen Elizabeth the 2nd crown.
- Added all the images for the King Charles the 3rd crown to the the public images folder sourced from govuk-frontend version 5.
- Added the headIcons block to the default layout to make sure the correct crown icons are used. 

## Type of change 🧩

- [ ] Documentation update


packages/forms-web-app/src/views/layouts/default.njk
```diff
@@ -4,6 +4,8 @@
 
 {%- set initiateScriptsConfig = [{ src: "/public/scripts/index.script.js" }] -%}
 
+{% set imagePath = '/public' %}
```
christopher-semmens, line 7: As this is the image path could we include the image folder: '/public/image' or change it to publicPath to accommodate manifest.json?

packages/forms-web-app/src/views/layouts/default.njk
```diff
@@ -12,6 +14,15 @@
 	{{ pageTitle }}
 {% endblock %}
 
+{# Note this block can be removed as well as the relevent images in /public/images when updating to v5 of the govuk-frontend node module #}
```
christopher-semmens, line 17: Just a suggestion but might be nice to put this chunk into an includes named 'head-icons.njk'.

packages/forms-web-app/src/views/layouts/default.njk
BenSurgisonPINS: Yes, I think I'll go with your second suggestion

packages/forms-web-app/src/views/layouts/default.njk
BenSurgisonPINS: To be honest, I think it's more obvious here with the comment that it needs to be removed when the upgrade to govuk-frontend V5 is implemented.  The additional file would need to be deleted.  I will implement as you have described if you think it still makes sense for it to be in a separate file.

christopher-semmens (approved):

PR #1328, opened 2024-09-04
test: update test to clear obj values and compare

christopher-semmens:
## Describe your changes

https://pins-ds.atlassian.net/browse/APPLICS-603?atlOrigin=eyJpIjoiMDc0ZWE3OTMyYTUxNGYzMTg2NjA2NmEzMGQwMDEyMGEiLCJwIjoiaiJ9

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/packages/forms-web-app/__tests__/setupTests.js
+++ b/packages/forms-web-app/__tests__/setupTests.js
@@ -4,14 +4,21 @@ const fetchMock = require('jest-fetch-mock');
 
 fetchMock.enableMocks();
 
+const clearAllObjValues = (translations) =>
+	Object.keys(translations).forEach((key) => {
+		if (typeof translations[key] === 'object') {
+			clearAllObjValues(translations[key]);
+		} else translations[key] = '';
+	});
+
 function toHaveSameKeys(a, b) {
-	const aKeys = Object.keys(a).sort();
-	const bKeys = Object.keys(b).sort();
+	clearAllObjValues(a);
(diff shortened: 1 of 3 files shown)


packages/forms-web-app/__tests__/setupTests.js
```diff
@@ -4,14 +4,21 @@ const fetchMock = require('jest-fetch-mock');
...
+		if (typeof translations[key] === 'object') {
+			clearAllObjValues(translations[key]);
+		} else translations[key] = '';
+	});
+
 function toHaveSameKeys(a, b) {
-	const aKeys = Object.keys(a).sort();
-	const bKeys = Object.keys(b).sort();
+	clearAllObjValues(a);
+	clearAllObjValues(b);
```
HadrianHughesKC, line 16: This is a nice solution, but it might have unintended side effects on other tests that run after the `.toHaveSameKeys` assertion because the underlying object is being mutated.

You could instead have `clearAllObjValues` construct a new object. Something like:

```javascript
const clearAllObjValues = (translations) =>
  Object.keys(translations).reduce((acc, key) => {
    if (typeof translations[key] === 'object') {
      acc[key] = clearAllObjectValues(translations[key]);
    } else {
      acc[key] = '';
    }

    return acc;
  })
```

HadrianHughesKC (changes requested):

HadrianHughesKC (approved):

PR #1386, opened 2024-11-25
feat: improve ni redirect setup process

Yakitrak:
## Describe your changes

Ticket: https://pins-ds.atlassian.net/browse/APPLICS-1142

PR improves the setup NI redirect process. The cloudflare workers already setup and working - this change uses the existing `BACK_OFFICE_API_INTEGRATION_CASE_REFERENCES` environment variable to enable cases to redirect instead of a hardcoded value on the JSON config. This will mean we can simply do a pipeline run to create a JSON config instead of having to do a code change, PR and merge like we currently do. Note, the cases in `BACK_OFFICE_API_INTEGRATION_CASE_REFERENCES` at the time of deployment will be used, this is not using feature flag on each redirect. 

## Type of change 🧩

- [ ] Documentation update


packages/ni-redirects/src/index.js
```diff
@@ -21,20 +27,25 @@ export default {
 	async fetch(request, env) {
 		/** @type {EnvConfig} */
 		const config = env.CONFIG === 'prod' ? prodConfig : testConfig;
+		/** @type {EnvGeneratedConfig} */
+		const generatedConfig = env.CONFIG === 'prod' ? prodGeneratedConfig : testGeneratedConfig;
+		const redirectAllowedCaseReferences = generatedConfig.redirectAllowedCaseReferences;
+		console.log(`Enabling redirection for the following cases: ${redirectAllowedCaseReferences}`);
```
beejjacobs, line 33: Maybe only log in test? This will use up precious CPU cycles!

azure-pipelines-deploy-ni-redirects.yml
```diff
@@ -34,6 +43,35 @@ jobs:
           steps:
             - download: none
             - checkout: self
+            - script: |
+                sudo apt-get update
+                sudo apt-get install -y jq
```
beejjacobs, line 48: If we need these tools ideally we get them built into the agent image, but we can handle that after we know its working.

beejjacobs (commented):
This seems like a pretty neat solution!

packages/ni-redirects/src/index.js
Yakitrak: Thanks good shout, changed to only log for test

azure-pipelines-deploy-ni-redirects.yml
Yakitrak: Sounds good, happy to remove stage when we add to image

beejjacobs (approved):

PR #1413, opened 2025-01-29
fix(api): handle error caused by querying ni projects table with utf chars

JamesP-KC:
## Describe your changes

APPLICS-1284 

The NI `wp_ipc_projects` table uses `latin1` encoding which means queries cannot include non-latin1 characters, and when doing so the database throws an errors, which the API does not handle and crashes.

This PR updates the NI project repository to return an empty result rather than throwing an error if a query with unsupported characters is received.

The `asyncRoute` middleware has also been added to the router/controller, so in the future any uncaught errors are caught at the router level rather than crashing the whole application.

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/packages/applications-service-api/src/repositories/project.ni.repository.js
+++ b/packages/applications-service-api/src/repositories/project.ni.repository.js
@@ -3,6 +3,7 @@ const { Op } = require('sequelize');
 const { pick } = require('lodash');
 const config = require('../lib/config');
 const { english: stopWordList } = require('../utils/stopwords');
+const logger = require('../lib/logger');
 const getApplication = async (caseReference) =>
 	db.Project.findOne({ where: { CaseReference: caseReference } });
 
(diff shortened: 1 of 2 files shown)


Yakitrak (approved):

emil-placheta (dismissed):

packages/applications-service-api/src/repositories/project.ni.repository.js
```diff
@@ -57,12 +57,22 @@ const getAllApplications = async (options = {}) => {
...
-	return { applications: rows, count };
+	try {
+		const { rows, count } = await db.Project.findAndCountAll({
+			...findAllOptions,
+			raw: true
+		});
+
+		return { applications: rows, count };
+	} catch (e) {
+		console.error(e);
```
DavidJWhiteDev, line 69: should we use something like logger.error() rather than console ?

DavidJWhiteDev (approved):
nice work James!

PR #1437, opened 2025-04-24
fix(web): localise CBOS exam timetable dates

emil-placheta:
https://pins-ds.atlassian.net/browse/APPLICS-1508

Issue is caused by the FO not handling dates correctly. 
CBOS is saving dates in UTC. In a situation when a timetable event (eg meeting or hearing) is created without the end time, as they usually are, a default time is set eg `2025-04-24 00:00:00.000` then prisma correctly saves the date in UTC and during summer time, subtracts an hour, saving the date in the db as 
`2025-04-23 23:00:00.000`. 
Not only does FO currently not handle timezones/DST at all and simply displays given values, it also ignores the time on the deadlines/meetings, and sets the time to end of day by default on all timetable dates. Changing CBOS date into `2025-04-23 23:59:59.000`.  Which in this situation makes the event timetable items display as a day early.

This fix simply adjusts the date before it is passed to the date handling logic, but only for the CBOS cases. 
Ideally, after the migration, we would remove this and keep the date logic using UCT, and format the display date only in the views.

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/packages/forms-web-app/src/pages/projects/examination-timetable/_utils/events/get-events.js
+++ b/packages/forms-web-app/src/pages/projects/examination-timetable/_utils/events/get-events.js
@@ -23,7 +23,11 @@ const getEvents = async (appData, i18n) => {
 			return {
 				...timetable,
 				dateOfEvent: localiseDate(timetable.dateOfEvent),
-				dateTimeDeadlineStart: localiseDate(timetable.dateTimeDeadlineStart)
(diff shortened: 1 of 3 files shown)


DavidJWhiteDev (approved):

DavidJWhiteDev (approved):
good spot on the nulls!

PR #1471, opened 2025-06-26
chore: set 'user: root' for azure-sql-edge to fix running in WSL

beejjacobs:
'mssql' container, running azure-sql-edge fails to start on WSL with an error around a 'master.mdf' file with a Windows path.

Adding `user: root` solves this, and aligns with other projects using this image.


ypk (approved):
LGTM 👍🏻

PR #1490, opened 2025-07-30
Data quality guide for all projects download

JanPhillips:
[Data quality guide for all projects download](https://pins-ds.atlassian.net/browse/APPLICS-1620)

## Describe your changes

- URL to the data quality guide is appended to .csv file for all projects.
- data quality guide spreadsheet is available on blob store for all environments.
- URL to the data quality guide spreadsheet changes according to runtime environment, between testing and prod. When in prod then url to Blob store for prod will be used, else url for Blob store for testing will be used.

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/packages/applications-service-api/__tests__/unit/utils/map-applications-to-csv.test.js
+++ b/packages/applications-service-api/__tests__/unit/utils/map-applications-to-csv.test.js
@@ -1,96 +1,141 @@
-const mapApplicationsToCsv = require('../../../src/utils/map-applications-to-csv');
-
-describe('mapApplicationsToCsv', () => {
-	it('returns a csv string', () => {
-		const applications = [
-			{
-				CaseReference: 'EN000001',
-				ProjectName: 'Test Project',
-				PromoterName: 'Test Promoter',
-				Proposal: 'Test Proposal',
-				Region: 'Test Region',
-				ProjectLocation: 'Test Location',
-				AnticipatedGridRefEasting: 'Test Easting',
-				AnticipatedGridRefNorthing: 'Test Northing',
-				LongLat: ['Test Long', 'Test Lat'],
-				Stage: 1,
-				Summary: 'Test Summary',
-				AnticipatedDateOfSubmission: 'Test Date Of Submission',
-				AnticipatedSubmissionDateNonSpecific: 'Test Submission Date Non Specific',
-				DateOfDCOSubmission: 'Test Date Of DCO Submission',
-				DateOfDCOAcceptance_NonAcceptance: 'Test Date Of DCO Acceptance Non Acceptance',
-				ConfirmedStartOfExamination: 'Test Confirmed Start Of Examination',
-				DateTimeExaminationEnds: '2024-01-01 00:00:00',
-				DateOfRecommendations: 'Test Date Of Recommendations',
-				ConfirmedDateOfDecision: 'Test Confirmed Date Of Decision',
-				DateProjectWithdrawn: 'Test Date Project Withdrawn'
-			},
-			{
-				CaseReference: 'EN000001',
-				ProjectName: 'Test Project',
-				PromoterName: 'Test Promoter',
-				Proposal: 'Test Proposal',
-				Region: 'Test Region',
-				ProjectLocation: 'Test Location',
-				AnticipatedGridRefEasting: 'Test Easting',
-				AnticipatedGridRefNorthing: 'Test Northing',
-				LongLat: ['Test Long', 'Test Lat'],
-				Stage: 2,
-				sourceSystem: 'back-office-applications',
-				Summary: 'Test Summary',
-				AnticipatedDateOfSubmission: 'Test Date Of Submission',
-				AnticipatedSubmissionDateNonSpecific: 'Test Submission Date Non Specific',
-				DateOfDCOSubmission: 'Test Date Of DCO Submission',
-				DateOfDCOAcceptance_NonAcceptance: 'Test Date Of DCO Acceptance Non Acceptance',
-				ConfirmedStartOfExamination: 'Test Confirmed Start Of Examination',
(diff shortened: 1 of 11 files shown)


packages/applications-service-api/src/utils/construct-url-for-blob-store-docs.js
ypk, line 1: here's another way you could write this


```
const ENVIRONMENTS = {
  prod: 'https://nsip-documents.planninginspectorate.gov.uk/',
  test: 'https://back-office-applications-docs-test.planninginspectorate.gov.uk/',
  dev: 'https://back-office-applications-docs-dev.planninginspectorate.gov.uk/',
  training: 'https://back-office-applications-docs-train.planninginspectorate.gov.uk/',
};

const DEFAULT_ENVIRONMENT = 'dev';

/**
 * Returns the base URL for blob store docs depending on the environment.
 * @param {string} [environment=process.env.NODE_ENV] - The environment string.
 * @returns {string} - The constructed URL.
 */
function constructUrlForBlobStoreDocs(environment = process.env.NODE_ENV) {
  return ENVIRONMENTS[environment] || ENVIRONMENTS[DEFAULT_ENVIRONMENT];
}

module.exports = constructUrlForBlobStoreDocs;
```

emil-placheta: bingo. only thing I would add is a string normalisation toLowerCase on the env var, but that might be my paranoia talking.

@JanPhillips Also, I might not be familiar with the work, but why do we need 4 versions of the same file across all envs in the first place? Can't we just have one in prod and keep the same link? Is the file different based on the environment?

packages/applications-service-api/src/utils/map-applications-to-csv.js
```diff
@@ -51,6 +54,21 @@ const mapApplicationsToCSV = (applications) => {
 		'Date withdrawn': application.DateProjectWithdrawn
 	}));
 
+	// insert blank rows
+	mappedApplications.push({});
+	mappedApplications.push({});
```
Tudor-PINS, line 59: Little nitpicky, but could this be more concise?

`mappedApplications.push({}, {}); `

packages/applications-service-api/src/utils/construct-url-for-blob-store-docs.js
JanPhillips: @emil-placheta thank you for that. I am not sure if we need 4 links either, the file is the same across all envs, it seemed like the safest option for me, if you think we only need the prod link then I'm very happy to implement just that.

DavidJWhiteDev: However, this would mean that any future change to the document would automatically immediately be in Prod, without any checking in Dev, Test, etc.  I think we should keep a doc for each environment for that reason.

beejjacobs: _FYI if you do need these URLs, they are already passed to the api as an env var: `AZURE_BLOB_STORE_HOST`. Would should avoid hard-coding values like this which can be configured._

emil-placheta: I was looking for an env var holding this in the portal in the api app, but that's not where it comes from. Is there an additional layer of env vars that gets added?

beejjacobs: @emil-placheta / @JanPhillips - D'oh I was looking at back office oops! However, that does mean the value already exists in terraform and it should be trivial to add to front office if required.

JanPhillips: A lot of thanks to Medini for helping me make [the changes for infrastructure-environment](https://github.com/Planning-Inspectorate/infrastructure-environments/pull/1099)

packages/applications-service-api/__tests__/unit/utils/map-applications-to-csv.test.js
```diff
@@ -1,96 +1,136 @@
-const mapApplicationsToCsv = require('../../../src/utils/map-applications-to-csv');
+describe('mapApplicationsToCSV', () => {
+	beforeEach(() => {
+		process.env.AZURE_BLOB_STORE_HOST = 'https://nsip-documents.planninginspectorate.gov.uk/';
```
emil-placheta, line 3: Rather than mocking the env var directly, you should mock the config where this env var is being pulled to

packages/applications-service-api/src/utils/map-applications-to-csv.js
```diff
@@ -61,4 +75,18 @@ const mapApplicationsToCSV = (applications) => {
 	});
 };
 
-module.exports = mapApplicationsToCSV;
+// Constructs the full URL to the published Quality Data Guide document.
+function getUrlForQualityDataGuide() {
```
emil-placheta, line 79: could the `qualityGuideFilename` be passed in to the function directly as a parameter

emil-placheta (changes requested):
Looks good, one thing remaining is to add the `AZURE_BLOB_STORE_HOST` into the docker-compose file to make it available for the local env as well

packages/applications-service-api/src/utils/map-applications-to-csv.js
emil-placheta: this function still uses `encodedFilename` from the outer scope. Even though this is a small utility function in a small file, it's good practice for this type of functions to have data they use be passed in parameters, which will help keep the function reusable and avoid potential side effects

emil-placheta (changes requested):
It doesn't look like the changes to `docker-compose.yml` with the azure host value have been included

emil-placheta (approved):
LGTM. Well done

PR #1496, opened 2025-08-05
fix(web): add datepublished filter for migrated case

prashantproX:
## Describe your changes -  added datepublished filter logic for migrated cases in document.backoffice.repository.js to support filtering documents where datePublished is stored as nvarchar.  

    Issue ticket number and link-  APPLICS-1639  https://pins-ds.atlassian.net/browse/APPLICS-1639 

    Please include a summary of the change along with any relevant context.- migrated documents have datePublished stored as nvarchar, which caused previous filters to miss results for single-day queries.
the new logic ensures both migrated and regular documents are correctly filtered.
    List any dependencies that are required for this change.- no new dependencies required. uses existing prisma client and query structure.

## Useful information to review or test - test single-day datePublished filters as well 

    Add any useful information that will help the reviewer test your changes.
    This could include example payloads, steps to reproduce, etc. - only test migrated cases 

## Type of change 🧩

- [ ] Documentation update


ypk (dismissed):
LGTM 👍🏻

ypk:
Please add test cases

packages/applications-service-api/src/repositories/document.backoffice.repository.js
```diff
@@ -29,6 +29,35 @@ const getDocuments = async (query) => {
...
+				lte: `${query.datePublishedTo}T23:59:59.999Z`
+			};
+		} else if (query.datePublishedFrom) {
+			dateConditions.datePublished = {
+				gte: `${query.datePublishedFrom}T00:00:00.000Z`
+			};
+		} else if (query.datePublishedTo) {
+			dateConditions.datePublished = {
+				lte: `${query.datePublishedTo}T23:59:59.999Z`
+			};
```
beejjacobs, line 55: FYI these hard-coded time strings may cause an issue. Assuming 'datePublished' is stored in UTC, then there may be issues with timezones. For example if a document was published at midnight during BST, the date would be 23:00:00 the previous day in UTC.

prashantproX: Good catch! Migrated cases store timestamps with time components, so the hardcoded timezone strings would indeed cause issues. Updated to use date-only comparison instead. Thanks! 👍
image

ypk (approved):

PR #1514, opened 2025-09-01
chore: use the data-model enum values for message event types

beejjacobs:
## Describe your changes
Use the data-model definition for message event type.

## Issue ticket number and link
https://pins-ds.atlassian.net/browse/RDSD-146


emil-placheta (commented):
Is is also worth updating `packages/event-client/src/event-type.js` so the strings from there won't get used in the future?

emil-placheta (approved):

DavidJWhiteDev (approved):
lgtm!

PR #1713, opened 2026-04-29
chore(api): remove usage of FEATURE_GENERAL_S51_BO flag

tomxelliott:
## Describe your changes
Remove usage of FEATURE_GENERAL_S51_BO flag

https://pins-ds.atlassian.net/browse/IRP-74


emil-placheta (changes requested):
Looks good, but the flag also needs to be removed from the root docker-compose file

emil-placheta (approved):

PR #1733, opened 2026-05-12
fix(apps): remove pagination options in query string to return correct results

tomxelliott:


Tudor-PINS (approved):
tested on local and seems to work, code looks good. going to test on dev after merge.

PR #1736, opened 2026-05-12
fix(web): stagger document submission (idas-428)

emil-placheta:
TLDR; Stagger document submission processing, to not overwhelm CBOS and avoid timeouts.

CBOS had an issue with duplicate refs, which were fixed in https://github.com/Planning-Inspectorate/back-office/pull/3152, however, this has surfaced an ikssue with Prisma 7, which due to a [bug ](https://github.com/prisma/prisma/issues/28002)prevents setting up individual timeouts for a transaction, and timing out after 15s, which is a problem when 20 documents are being sent from FO and are processed in sequence (to avoid duplicate references) in CBOS.

## Type of change 🧩

- [ ] Documentation update


packages/forms-web-app/src/pages/examination/process-submission/utils/process.js
```diff
@@ -22,6 +24,10 @@ const handleProcessSubmission = async (session) => {
 				if (submissionId) form.append('submissionId', submissionId);
 				const response = await postSubmission(session.caseRef, form);
 				submissionId = response.data.submissionId;
+				//Due to a bug with Prisma 7, preventing from assigning individual timeouts, we stagger the document submissions
+				//to prevent CBOS from being overwhelmed and timing out
+				//https://github.com/prisma/prisma/issues/28002
+				await sleep(3000);
```
Tudor-PINS: I know it's more complex so this may not be necessary for a temporary change, however.

If we have 7 files to process, we only need 6 sleeps, however this process sleeps after each one meaning we sleep once unnecessarily.

It probably is so small that it's not an issue, but using a flatmap and doing n-1 sleeps would be the best implementation of this. What are you thoughts on if this is necessary?

emil-placheta: nice spot, I'll have a look

emil-placheta: moved the wait to the top of the loop and paired with the submission Id assignment. This way it will wait _before_ submission rather than after but because the submission id is not there on first iteration, as the submission has not yet been done, it does not wait on the first loop. 

So it subtracts from the loop count, but from the beginning, rather than the end (y)

Tudor-PINS (approved):

PR #1743, opened 2026-05-17
feat/idas-131 display boundary shapefiles

ypk:
## Describe your changes

[[IDAS-131] Display Project Boundary Shape Files - Project Information Page](https://pins-ds.atlassian.net/browse/IDAS-131)

Displays project boundary shapefiles on individual project information pages. The map now renders the project's GeoJSON boundary polygon (sourced from GIS shapefile data) and automatically fits the view to the full boundary.

When no shapefile is available, the existing pin view using grid references is shown as a fallback.

A message is displayed below the map in both English and Welsh indicating either that boundaries are illustrative

## Type of change 🧩

- [ ] Documentation update

[IDAS-131]: https://pins-ds.atlassian.net/browse/IDAS-131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cprieststephens:
I can see that the GeoJSON data for the individual project map is being accessed from applicationData in res.locals, and is accessed by the function that iterates over the applications array for the all projects map.  Is the back-end code to fetch the GeoJSON data from the blob store and add it to applicationData or return it with existing applications data being implemented in another ticket?

packages/forms-web-app/src/scripts/projects-map/layers.js
```diff
@@ -0,0 +1,225 @@
...
+}
+
+function buildGeoJsonLayer(map, features) {
+	const boundarySource = new VectorSource({ features });
+
+	const layer = new VectorLayer({
+		source: boundarySource,
+		style: new Style({
+			fill: new Fill({ color: 'rgba(0, 112, 184, 0.15)' }),
+			stroke: new Stroke({ color: '#0070B8', width: 2 })
```
cprieststephens: Should the polygon boundary and fill colour be red rather than blue?

ypk: good spot!, I will change this to red

ypk:
> I can see that the GeoJSON data for the individual project map is being accessed from applicationData in res.locals, and is accessed by the function that iterates over the applications array for the all projects map. Is the back-end code to fetch the GeoJSON data from the blob store and add it to applicationData or return it with existing applications data being implemented in another ticket?

I don't think that's part of this ticket 

@DeanH1990 what are your thoughts on this?

packages/forms-web-app/src/scripts/projects-map/layers.js
ypk: updated the colours and width of the stroke as per spec from @alexrehopkins 
image

packages/forms-web-app/src/pages/projects-map/controller.js
```diff
@@ -74,20 +25,32 @@ const getProjectsMapController = async (req, res, next) => {
...
+			application.geojson = new GeoJSONBuilder().addApplications([application]).build();
+			return application;
+		});
+
+		const geoJSON = {
+			type: 'FeatureCollection',
+			features: applicationsWithGeoJSON.flatMap((application) =>
+				Array.isArray(application?.geojson?.features) ? application.geojson.features : []
+			)
+		};
```
DeanH1990, line 52: I have a couple of questions here:

Where does the `.geojson` property come from? I followed the `getApplications` request through to the API mapping and couldn't see `.geojson` being added to the application payload.

Also, these changes appear to modify the all-projects map boundary behaviour by building a FeatureCollection in the web app. My understanding was that the all-projects boundary work is covered separately under IDAS-514 using the master geojson boundary file, so just wanted to check whether these changes are intentionally part of this ticket or if this is overlapping with that work.

packages/forms-web-app/src/pages/projects/index/controller.js
```diff
@@ -41,15 +42,23 @@ const getProjectsIndexController = async (req, res, next) => {
...
 				getRule6DocumentType(caseRef),
 				getRule8DocumentType(caseRef),
 				getMiscDataByStageName(applicationData.status.text, caseRef),
-				applicationData.longLat ? getMapAccessToken() : Promise.resolve(null)
+				hasMapData ? getMapAccessToken() : Promise.resolve(null)
 			]);
 
+		const mapGeoJSON = hasProjectBoundary
+			? JSON.stringify(applicationData.geojson)
+			: JSON.stringify(pointGeoJSON);
```
DeanH1990, line 61: Similarly here, where is `applicationData.geojson` populated?

From tracing the existing application API responses I couldn't see a `geojson` field being mapped, so just wanted to confirm this is already available

packages/forms-web-app/src/pages/projects-map/controller.js
ypk: the feature to fetch the geojson from the blob store and hand it to this controller is currently not available.
I am expecting the data to be pulled somewhere upstream and added to the response of `getApplications()` call which my controller calls. 

I was hoping that your work to pull the geojson from blob store would also add the geojson to the data so I can call  `getApplications()` and use it in individual maps

--

regarding this work encroaching on IDAS-514, it does not (for now), it enhances the existing functionality to fetch all records and creates geoJSON from `LatLng` in case of no `.geojson` prop available. All this mechanism is temporary and will be super seeded when master geojson comes into play. This is just to keep the existing map functioning as expected

Tudor-PINS (approved):
lgtm but please resolve open comments, some queries from dean that aren't answered

don't forget to squash too :)

Tudor-PINS (changes requested):
ignore approval, changes needed as per Dean's comment. reviewed wrong PR!

PR #1760, opened 2026-06-02
Feat/idas 131 display boundary using shapefiles

ypk:
## Describe your changes
This Implements loading shapefiles on projects/index

[\[IDAS-131\] Display Project Boundary Shape Files - Project Information Page](https://pins-ds.atlassian.net/browse/IDAS-131)

## Type of change 🧩

- [ ] Documentation update


packages/forms-web-app/src/pages/projects/index/controller.js
```diff
@@ -16,8 +17,38 @@ const {
 } = require('./_utils/examination-or-decision-completed-date');
 const { getMapAccessToken } = require('../../_services');
 const { isBackOfficeCaseReference } = require('./_utils/is-backoffice-case-reference');
+const { GeoJSONBuilder } = require('../../../lib/geojson-builder');
+const { getDocumentByType } = require('../../../lib/application-api-wrapper');
+const { projectsIndexViewPath, geoJsonMimeType } = require('./config');
 
-const view = 'projects/index/view.njk';
```
DeanH1990, line 20: Is there a reason view has been moved in to config.js?

We generally keep the view path in the controller, so Im trying to understand what advantage we gain by moving it into config.

packages/forms-web-app/src/scripts/projects-map/index.js
```diff
@@ -1,346 +1,187 @@
...
-			<table class="cluster-popup-table">${rows}</table>
-		</div>`
-	);
-}
-
-async function loadBoundaryLayer(map, fallbackVectorSource) {
+async function loadBoundaries(map, pointSource) {
 	try {
-		const response = await fetch(getGeoJsonURL());
+		const response = await fetch(BOUNDARY_GEOJSON_URL);
```
DeanH1990: Is there a reason were no longer using the util helper function here? 

The helper already encapsulates how the master boundary endpoint is constructed, whereas introducing the hardcoded const creates another way of obtaining the same endpoint, if the endpoint were to change in the future it would need updating in multiple places

DeanH1990 (commented):
This is quite a substantial change to the shared map implementation to support individual project boundaries. Could you give a brief overview of which parts of the refactor were required to implement the individual project boundaries vs general restructuring/ refactoring?

ypk:
> This is quite a substantial change to the shared map implementation to support individual project boundaries. Could you give a brief overview of which parts of the refactor were required to implement the individual project boundaries vs general restructuring/ refactoring?

I did give you  detailed walk though of each and every file and change in person on May 20th when we were in the office 😿 
I will schedule a call to go through this again

packages/forms-web-app/src/pages/projects/index/controller.js
ypk: fair enough, I will revert this.

The idea was to have all string variables in config or cleaner separation. no functional advantage

DeanH1990:
> > This is quite a substantial change to the shared map implementation to support individual project boundaries. Could you give a brief overview of which parts of the refactor were required to implement the individual project boundaries vs general restructuring/ refactoring?
> 
> I did give you detailed walk though of each and every file and change in person on May 20th when we were in the office 😿 I will schedule a call to go through this again

Thanks, I remember the walkthrough. I was mainly asking from a PR review perspective now that I can see the full implementation and associated tests.

packages/forms-web-app/src/lib/__tests__/test-utils.js
```diff
@@ -0,0 +1,104 @@
...
+
+function randomValidApp() {
+	return {
+		LongLat: [randomLng(), randomLat()],
+		CaseReference: randomString(5, 12),
+		ProjectName: randomString(3, 30),
+		Stage: Math.floor(Math.random() * 9)
+	};
+}
+
```
emil-placheta: Putting aside the whole testing approach using randomness. These two functions are identical?

packages/forms-web-app/src/lib/__tests__/geojson-builder.test.js
```diff
@@ -0,0 +1,382 @@
...
+			for (const { num, label } of stages) {
+				const builder = new GeoJSONBuilder();
+				const result = builder
+					.addApplications([
+						{ CaseReference: 'REF', ProjectName: 'P', Stage: num, LongLat: [0, 0] }
+					])
+					.build();
+				expect(result.features[0].properties.stage).toBe(label);
+			}
+		});
```
emil-placheta, line 114: I would avoid using loops within tests like this, as this makes it difficult to read test failures and harder to maintain. Using it.each would be more readable and provide better error messages when test fails

packages/forms-web-app/src/lib/__tests__/geojson-builder.test.js
emil-placheta, line 1: We should also probably nest all tests under one describe, (which can have additional describes within it) rather than have 3 separate ones, as this is the current pattern and makes it clear when reading results it all belongs to a single test suite, and not 3 separate ones.

packages/forms-web-app/src/lib/__tests__/test-utils.js
emil-placheta, line 1: rather than creating extra functionality and adding random  'data seeders', why not create static fixtures with invalid data, which tests can run against? This would make it easier to determine which values are considered valid vs invalid, and make tests more readable. It would also serve as a form of documentation for the feature, where reading the tests helps you understand how the functionality is meant to work

packages/forms-web-app/src/lib/__tests__/geojson-builder.test.js
ypk: I’ve replaced the assertions with `it.each` so each case is reported independently

packages/forms-web-app/src/lib/__tests__/geojson-builder.test.js
ypk: I’ve grouped everything under a single top-level GeoJSONBuilder describe, with nested describes for each method/behavior

packages/forms-web-app/src/lib/__tests__/test-utils.js
ypk: I refactored the code to remove the random data seeder and made the test data static, so the cases are easier to understand. I then moved that into the file and deleted the util file.

packages/forms-web-app/src/scripts/projects-map/index.js
ypk: I’ve switched this back to use the existing helper so the boundary endpoint is in one place

packages/forms-web-app/src/lib/__tests__/test-utils.js
ypk: That duplication slipped in during the test-helper refactoring. I’ve removed the duplicate seeder and simplified the tests to use fixtures

DeanH1990 (approved):

PR #1816, opened 2026-07-24
fix(web): increase express query parser array limit (idas-694)

DeanH1990:
## Describe your changes

On investigation Express uses the `qs` query parser, which has a default `arrayLimit` of 20. When more than 20 values are supplied for the same query parameter, they are parsed as an indexed object instead of an array.
This caused the document filters to be sent to the API in an invalid format, resulting in the validation error:
`'filters.0.type.0.value' must be string`

I have configured Express to use a custom `qs` query parser with an increased `arrayLimit` so filter values continue to be parsed as arrays regardless of how many are selected.
I initially considered normalising the values in makeIntoArray(), which fixed the API request but left the UI in an inconsistent state (selected filters and checkbox state were no longer preserved). Configuring the query parser resolves the issue at the source for all consumers of req.query.

This behaviour is documented by Express/qs and discussed here:
https://github.com/expressjs/express/issues/4053

[ticket 694](https://pins-ds.atlassian.net/jira/software/c/projects/IDAS/boards/1034?selectedIssue=IDAS-694)

Locally there is a test case with ref EN010116 which has 46 filters under the pre-application stage and reproduced the bug. With this solution the bug is no longer present, you can change the arrayLimit override to less 45 and select all the filters and it will cause the issue again.

## Type of change 🧩

- [ ] Documentation update


packages/forms-web-app/src/app.js
```diff
@@ -38,6 +39,13 @@ const { attachCorrelationId } = require('./middleware/attach-correlation-id');
 
 const app = express();
 
+app.set('query parser', function (str) {
+	return qs.parse(str, {
+		allowPrototypes: true,
+		arrayLimit: 50
+	});
```
DeanH1990, line 46: The production case affected by this issue currently has 21 document filters under the Examination stage, which exceeds the default qs arrayLimit of 20. This is why removing any one filter causes the request to succeed.

I'm happy to reduce this value from 50 if there's a preferred limit.

cprieststephens (approved):
LGTM

